PR #84, opened 2020-11-10
feat: deploy every day

juzhiyuan:
Deploy the docs from apisix once a day

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,6 +7,9 @@ name: CI
 on:
   push:
     branches: [ master ]
+  schedule:
+  # Run everyday at 9:00 AM (See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07)
+  - cron: "0 9 * * *"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -7,6 +7,9 @@ name: CI
 on:
   push:
     branches: [ master ]
+  schedule:
+  # Run everyday at 9:00 AM (See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07)
+  - cron: "0 9 * * *"
```
LiteSun, line 12: What time zone is this in?

juzhiyuan: UTC of course.

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#schedule

LiteSun: > UTC of course.
> 
> https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#schedule

Got it.

LiteSun (approved):

membphis (approved):

PR #123, opened 2020-12-31
fix: the download links of  asc and sha512 should use https://downloa…

gxthrj:
Fix: the download links of  asc and sha512 should use https://downloads.apache.org/apisix/apisix-ingress-controller-0.1.0/...

Related to https://github.com/apache/apisix-ingress-controller/issues/146


membphis (approved):

tokers (approved):

sebbASF:
Sorry, but this is also wrong.
Release candidates (e.g. RC2) are not intended for release.
There is no need to change the file names, just the paths to the hash and sig.

gxthrj:
> Sorry, but this is also wrong.
> Release candidates (e.g. RC2) are not intended for release.
> There is no need to change the file names, just the paths to the hash and sig.

Fixed in #125

PR #141, opened 2021-01-15
upgrade: docusaurusV2

KishaniKandasamy:
Changes:
upgrade: docusaurusV2

Screenshots of the change:
Screenshot (19)


KishaniKandasamy:
Screenshot (17)
I'm stuck at this point  as **CompLibrary is deprecated in v2**. Therefore when I move pages/en files in v1 to src/pages in v2, It shows errror like "Cannot find module: CompLibrary".

juzhiyuan:
Great! We will review recently :)

website2.0/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "browserslist": {
+    "production": [">0.5%", "not dead", "not op_mini all"],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  }
+}
```
juzhiyuan, line 30: We need a new line here :)

website2.0/static/css/custom.css
```diff
@@ -0,0 +1,23 @@
+/**
```
juzhiyuan, line 1: After we migrated all needed files, we need to add ASF header here.

.gitignore
```diff
@@ -1,12 +1,28 @@
+# dependencies
+website2.0/node_modules
```
juzhiyuan, line 2: Hi, could we omit the wesite2.0 directory? Set all needed files under the root directory directly is fine.

juzhiyuan:
I see there has a cli to help us migrate.

https://v2.docusaurus.io/docs/migration/automated

KishaniKandasamy:
> I see there has a cli to help us migrate.
> 
> https://v2.docusaurus.io/docs/migration/automated

Yes. Initially I did  running the migration cli, and tried to complete the missing parts  using  the manual migration process.

# migration command format
[1]npx @docusaurus/migrate migrate ( v1 website directory)  (desired v2 website directory)
Ex: npx @docusaurus/migrate migrate ./v1-website ./v2-website

That's why I created website2.0 :)
Screenshot (20)

But here It says **some parts can't easily be automated, and  have to fallback to the manual process.**

website2.0/package.json
KishaniKandasamy: Yes .I will correct it :)

website2.0/static/css/custom.css
KishaniKandasamy: yes Juzhiyuan

juzhiyuan:
ok, just to make sure, this PR is created by the following steps:

1. using the command to auto migrate;
2. fix missing parts which that command cannot do.

.gitignore
KishaniKandasamy: Yes Juzhiyuan .It would be better to under have the root directory   .

KishaniKandasamy:
> ok, just to make sure, this PR is created by the following steps:
> 
> 1. using the command to auto migrate;
> 2. fix missing parts which that command cannot do.

Yes

KishaniKandasamy:
Here the problem is Some parts of the migration can't be automated (**particularly the pages**), so we will have to migrate them manually.
Also  CompLibrary is deprecated in v2. Therefore when I move pages/en files in v1 to src/pages in v2, It shows errror like "Cannot find module: CompLibrary".
The guide says to write our own React component or use Infima styles.So **First Do we need to modify the  pages according to this** ?
Screenshot (22)

juzhiyuan:
Hi @KishaniKandasamy, I would prefer using Infima styles, but which one is easier for you, then we use it.

KishaniKandasamy:
> Hi @KishaniKandasamy, I would prefer using Infima styles, but which one is easier for you, then we use it.

Hi Zhiyuan , Yes. If you give me sometime, I could  solve this using Infima styles. as I am not most  familiar with Infima styles.

juzhiyuan:
Sure :)

PR #162, opened 2021-02-04
fix: update CD configuration

### LINKED ISSUE #157: FIx auto deploy
Due to we upgrade docusaurus from V1 to V2, our CD[1] is broken but not hard to fix.

[1] https://github.com/apache/apisix-website/blob/master/.github/workflows/deploy.yml

juzhiyuan:
Fixes: #157 

Changes:

Update CD configuration to use correct publish dir, omit apisix's docs first.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,8 +7,6 @@ name: CI
 on:
   push:
     branches: [master]
-  pull_request:
-    branches: [master]
   schedule:
     # Run everyday at 9:00 AM (See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07)
     - cron: "0 9 * * *"
@@ -30,6 +28,7 @@ jobs:
 
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'	
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: website/build
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,7 +29,7 @@ jobs:
         run: |
           git clone https://github.com/apache/apisix.git
           cp -r apisix/doc/images ./website/static && rm -rf apisix/doc/images
-          cp -r apisix/doc ./docs/apisix
+          cp -r apisix/doc ./website/docs/apisix
 
       - name: Build
         run: cd website && yarn && yarn build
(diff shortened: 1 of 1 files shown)


imjoey:
@juzhiyuan  The `docs` directory has also been moved to `website`, so please use `cp -r apisix/doc ./website/docs/apisix` instead of `cp -r apisix/doc ./docs/apisix` in step **Clone apisix's docs from repo & update image's link**

.github/workflows/deploy.yml
```diff
@@ -23,30 +23,13 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
-      - name: Clone apisix's docs from repo & update image's link
```
imjoey, line 26: @juzhiyuan may I ask the reason to remove APISIX docs? They are useless, or have other alternatives?

juzhiyuan: because currently, docs from `apisix/doc` need to be adjusted in many places(like images position) to have a better view (not good though), will add them back.

https://github.com/apache/apisix-website/pull/162/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3L35-L45

imjoey: ok. Thanks for the tip.

imjoey (approved):
/lgtm

.github/workflows/deploy.yml
```diff
@@ -23,30 +23,13 @@ jobs:
...
-          find . -type f -name "*.html" -print0 | xargs -0 sed -i -e 's/src="\.\/images/src="https:\/\/apisix\.apache\.org\/images/g'
-          find . -type f -name "*.html" -print0 | xargs -0 sed -i -e 's/src="images/src="https:\/\/apisix\.apache\.org\/images/g'
-
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
         if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: website/build/apisix-website
+          publish_dir: website/build
```
liuxiran, line 34: we will publish the yarn build results right? I tried to find this cinfig but failed, so may I ask if we have configuration for this address in any config files or it is a default yarn build localcation? thanks

juzhiyuan: ya, it's the default output folder.

liuxiran: got it, thanks

liuxiran (approved):

PR #172, opened 2021-02-09
docs: remove redundant step in release-guide.md

starsz:

Changes:

Hello, now the apisix and apisix-dashboard project's `Makefile` will exclude the `.git` file.
So I think we don't do this step in the release-guide.

See: https://github.com/apache/apisix-dashboard/blob/fda123c28dda7e4f4ae5d649b286ffcaebc568ae/Makefile#L114
See: https://github.com/apache/apisix/blob/c895cdf4ff008c13dec8197e9f0a0407be144517/Makefile#L226

Screenshots of the change:

### DIFF
--- a/website/docs/release-guide.md
+++ b/website/docs/release-guide.md
@@ -208,9 +208,6 @@ git clone -b v$APISIX_VERSION git@github.com:apache/apisix.git apache-apisix-$AP
 # check version
 $ cd apache-apisix-$APISIX_VERSION && ./utils/check-version.sh $APISIX_VERSION && cd ..
 
-# delete .git
-$ rm -rf apache-apisix-$APISIX_VERSION/.git
-
 # make tar package / asc / sha512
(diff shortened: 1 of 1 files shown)


juzhiyuan (approved):

LiteSun (commented):
It is possible that the user may not pack through the makefile, so to be on the safe side I don't think we should remove line211-213. cc @nic-chen

nic-chen:
> It is possible that the user may not pack through the makefile, so to be on the safe side I don't think we should remove line211-213. cc @nic-chen 

agree with you,we could keep them

tokers (approved):

starsz:
> It is possible that the user may not pack through the makefile, so to be on the safe side I don't think we should remove line211-213. cc @nic-chen

Because the line 214 ~ 215 tell us to use`make release`.

PR #187, opened 2021-02-13
feat: add Dependabot config file and configure updates to Actions eve…

jbampton:
…ryday

Fix missing HTTPS on link.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
```
juzhiyuan, line 6: Just to make sure, this is going to check GitHub Action version everyday?

jbampton: Yes it will check for updates to `GitHub Actions` daily.  But it does not mean it will update everyday.

juzhiyuan (approved):

tokers (approved):

PR #205, opened 2021-02-25
feat: sync docs & deploy

juzhiyuan:
sync docs first then deploy

Please see http://apisix.apache.org/docs/ingress-controller/design


sync-docs.js
```diff
@@ -38,7 +39,10 @@ const replaceMDImageUrl = (project, paths) => {
         .match(/\((.+?)\)/g)[0]
         .replace("(", "")
         .replace(")", "")
-        .replaceAll("../", "");
+        .replace("../", "")
+        .replace("../", "")
+        .replace("../", "")
+        .replace("../", "");
```
Yiyiyimu, line 45: I noticed this change is in https://github.com/apache/apisix-website/pull/205/commits/eda04d98b37a25dbd8dab5383b7e6372256c69fc to pass CI, but it seems a bit hacky and I can't get what's the meaning of this 🤣

juzhiyuan: it's a trick here, to replace some paths like `../../../xxxx`, `../../xxxx`

juzhiyuan: because replaceAll function is not available here, so I use a trick here. This JS file will get improved in the future.

.github/workflows/deploy.yml
```diff
@@ -9,7 +9,7 @@ on:
     branches: [master]
   schedule:
     # Run everyday at 9:00 AM (See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07)
```
LiteSun, line 11: What timezone is used here?

LiteSun (approved):

.github/workflows/deploy.yml
juzhiyuan: GMT+8

gxthrj (approved):

PR #272, opened 2021-03-27
feat: Add a bundle of Good first issues links to Homepage

### LINKED ISSUE #178: Add a bundle of Good first issues links to Homepage
Add a bundle of Good first issues Section to the Homepage.

[1] https://github.com/apache/apisix#community

KishaniKandasamy:
Fixes: #178

Changes:

update homepage

Screenshots of the change:

Screenshot (64)

### DIFF
--- a/website/src/css/customTheme.css
+++ b/website/src/css/customTheme.css
@@ -503,18 +503,40 @@ a:hover {
   display: flex;
 }
 
-/*newsletter section*/
+/*contribution section*/
+.contribution {
+  justify-content: center;
+  margin-bottom: 100px;
+}
 
-.center-elem{
+.center-elem {
   display: flex;
   justify-content: center;
 }
 
-.news-logo{
+.contribution-text {
+  color: var(--color-primary);
+  text-align: center;
+}
+
+.contribution-links {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+/*newsletter section*/
+
+.center-elem {
+  display: flex;
+  justify-content: center;
+}
+
+.news-logo {
   margin-bottom: 20px;
 }
 
-.news-button{
+.news-button {
   border: #111827;
   padding: 5px 30px;
   border-style: solid;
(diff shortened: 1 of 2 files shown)


juzhiyuan:
Hi, I'm not sure if we need those 3 buttons 😂 need @qier222 @1502shivam-singh 's confirm

KishaniKandasamy:
> Hi, I'm not sure if we need those 3 buttons 😂 need @qier222 @1502shivam-singh 's confirm

1502shivam-singh:
I too feel the same, no need for 3 buttons. A "View on GitHub" is in most cases enough, we have a **contribution guidelines** section to handle too. So, making a "View on Github" as primary and "See guidelines" button as secondary would do. 
**We can try hyperlinks too** if buttons seem an overkill. 
If you are going to use buttons, check that first the primary one comes, then the secondary.

1502shivam-singh:
> > Hi, I'm not sure if we need those 3 buttons 😂 need @qier222 @1502shivam-singh 's confirm
> 
> 

The key is the presentation, check that you have used buttons, while there they have used hyperlinks.
Buttons seem an overkill and almost intrusive for this cause. More of a problem with UI choice, might try some different UI pattern to make it look good while still using buttons instead of a simple center aligned section.
Three buttons also lead to UX problems, like too many choices, overwhelming for users (Depends on visual arrangment). While hyperlinks don't as they are smaller in size and lesser overwhelming than buttons.

KishaniKandasamy:
> > > Hi, I'm not sure if we need those 3 buttons 😂 need @qier222 @1502shivam-singh 's confirm
> > 
> > 
> > 
> 
> The key is the presentation, check that you have used buttons, while there they have used hyperlinks.
> Buttons seem an overkill and almost intrusive for this cause. More of a problem with UI choice, might try some different UI pattern to make it look good while still using buttons instead of a simple center aligned section.
> Three buttons also lead to UX problems, like too many choices, overwhelming for users (Depends on visual arrangment). While hyperlinks don't as they are smaller in size and lesser overwhelming than buttons.

Okay Thanks for the feedback

PR #289, opened 2021-04-08
chore: support dark mode and switch

Yiyiyimu:

Screenshots of the change:

image

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -314,7 +314,9 @@ module.exports = {
       contextualSearch: true,
     },
     colorMode: {
-      disableSwitch: true,
+      defaultMode: 'light',
+      disableSwitch: false,
+      respectPrefersColorScheme: true,
     },
     metadatas: [
       {


Yiyiyimu:
An awkward thing is when text is not in black (for example some dark grey), dark mode seems could not get its opponent color...

image

One possible solution to deal with this situation is like the tagline of https://docusaurus.io/, to make it stale and not get affected by the dark mode.

**Light mode**
image

**Dark mode**
image

I tried to find the method to make it stale but failed 😿 crying again for not familiar with frontend

jbampton:
I think for sure it's like JS/CSS/HTML but I suck at front end too.  
You may have seen some of my sites it's all weird purple disco.
But for sure dark mode is important so we need this

juzhiyuan (approved):
Though it's wired due to some color issues, we still need this.

tokers (approved):

Yiyiyimu:
> Though it's wired due to some color issues, we still need this.

I'll create a new issue for this

PR #293, opened 2021-04-09
chore: disable version since we do not support it

Yiyiyimu:

Remove the useless version switcher

image


Yiyiyimu:
Couldn't find methods to disable it. Close and leave it to issue

PR #299, opened 2021-04-10
fix: disable version switch button in sidebar, as website doesn't support it

### LINKED ISSUE #294: bug: need to disable version button since we do not support it
Currently we got a version switch on docs, but actually, we don't need it since we do not support multi-version docs, so it's better to remove it.

image

An example could be docusaurus page.

image

1502shivam-singh:
Fixes: #294

Changes:
Deleted the version switch button in sidebar of docs

Screenshots of the change:

image


juzhiyuan (approved):

juzhiyuan:
ok, we could disable it for now.

1502shivam-singh:
TBH, I was also quite reluctant to do this, given we might have to add versioning sometime later. But anyway, could add it later when actually versioning the docs, because for now it serves no purpose.

LiteSun (approved):

PR #297, opened 2021-04-10
fix: darkMode enhanced

iamayushdas:
Signed-off by: Ayush Das (ayush24das@gmail.com)

Fixes: #291 

Changes:

Dark mode fixed and enhanced

Screenshots of the change:

[Website change 1](https://user-images.githubusercontent.com/40708551/114260533-16968d00-99f3-11eb-9c94-3da6bf478f4f.png)

PR #303, opened 2021-04-12
chore: add alpha channel to showcase image for dark mode

Yiyiyimu:

related to #297

Scripts thanks to https://stackoverflow.com/questions/35859140/remove-transparency-alpha-from-any-image-using-pil

1. convert svg to png, to make it possible to add background
2. mask all images alpha channel, so the background would be white instead of transparency.


Yiyiyimu:
Since the images coming from CDN, so the preview sites could not show updated images. It seems the only thing I could do is to add some screenshots of opening them locally in photoshop 🤦 

Before:
image

After:
image

juzhiyuan:
image
image

Wow, I see so many images got updated, but some images seem broken or too dark?

Yiyiyimu:
> Wow, I see so many images got updated, but some images seem broken or too dark?

@juzhiyuan On my side, no images got updated 🤣 The reason is all images coming from CDN, and it would not got updated when only in PR. 

That's the same reason those two images got broken. I changed them from svg to png, so we could not use the original source from CDN anymore.

juzhiyuan (approved):

membphis (approved):

PR #334, opened 2021-05-07
docs: fix  edit links would be broken for zh doc page while repo not support zh doc

### LINKED ISSUE #332: bug: doc edit links would be broken for zh doc page while repo not support zh doc
> I tried to search those links in every repositories, but those links doesn't exist now. @iamayushdas could you check the issue once again.

https://user-images.githubusercontent.com/40708551/116728234-4853b000-aa03-11eb-92b7-09c5064e5f2e.png

The reason those links are broken is due to, apisix-ingress-controller does not contain Chinese doc, but on Chinese website, the edit link is pointing to `zh` link. So the logic of multi-language doc edit links needs to fix. But we could leave it to another PR.

_Originally posted by @Yiyiyimu in https://github.com/apache/apisix-website/issues/328#issuecomment-833172368_

KishaniKandasamy:
Fixes: #332

Changes:
update editurl  function

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -151,7 +151,7 @@ module.exports = {
           docPath,
           permalink,
         }) {
-          return `https://github.com/apache/apisix-dashboard/edit/master/docs/${locale}/latest/${docPath}`;
+          return `https://github.com/apache/apisix-dashboard/edit/master/docs/en/latest/${docPath}`;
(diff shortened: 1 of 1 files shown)


Yiyiyimu (approved):
Thanks for the contribution!

juzhiyuan (approved):

PR #553, opened 2021-09-04
feat: optimize the blog page display effect

Baoyuantop:
Changes:

By modifying the content of `BlogListPage` `BlogPostItem` `BlogSidebar` `BlogPostPage` `BlogTagsPostsPage` in `@docusaurus/theme-classic`.

Screenshots of the change:

image

image

image

What do you think of it? You can see the effect through the preview link, welcome to point out problems or make suggestions ～


1502shivam-singh:
> By modifying the content of `BlogListPage` `BlogPostItem` `BlogSidebar` `BlogPostPage` `BlogTagsPostsPage` in `@docusaurus/theme-classic`.

Looks really clean. Now that these components are swizzled we are in full control of our blog.

> What do you think of it? You can see the effect through the preview link, welcome to point out problems or make suggestions ～

Something I noticed, when I click on, say `technology` tag then `security` tag disappears. Similarly for others, if I click some tag than some other tag disappears in sidebar.
So if I am correct, you are following some graph type structure here, like on clicking on `technology` it will also show other tags that come with blogs that contain `technology` tag. 

This, as is understandable can become little confusing. So I guess an option could be keeping the sidebar static, not changing on clicking on tags, just changing the background of the tag clicked to red.

What do you think ? Or was the current effect expected for some reason ?

website/src/theme/BlogSidebar/index.js
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  return (
+    <div className={clsx(styles.sidebar, 'thin-scrollbar')}>
+      <h3 className={styles.sidebarItemTitle}>Tags</h3>
+      <div className={styles.sidebarItemList}>
+        {Object.entries(count).map(([tag, num]) => (
+          <div
+            key={tag}
+            className={`${styles.sidebarItem} ${selected === tag ? styles.selected : ''}`}
```
1502shivam-singh, line 55: Here, you might want to check the comparison `selected === tag` especially for the `practical case` tag.
As, when clicking on practical case tag, it's background color doesn't hold. 
My guess for this bug is, you are probably comparing "practical case" and "practical-case" or something similar here.

liuxiran:
Awesome, thanks @Baoyuantop .

and there are something would make this feature better to use, as @1502shivam-singh pointed in https://github.com/apache/apisix-website/pull/553#issuecomment-913048731
> This, as is understandable can become little confusing. So I guess an option could be keeping the sidebar static, not changing on clicking on tags, just changing the background of the tag clicked to red.

this suggestion would be a better choice, 

and more details :
* when user choose a tag, e.g. technology,  this tag should be marked color red, and all the posts contains technology tag should be listed.
* when user choose two tags, e.g. technology and news, these two tags should be marked color red, and all the posts contains technology should be list, and all the posts contains news tag should be list too. Attention that, the posts contains both  technology and news tags, should be only list one time.
* when user choose `All` tag, clear all other tags selected, and list all posts.

what do you think?

Baoyuantop:
I found that my problem lies in the fact that the data of the tags on the left depends on the list data on the right instead of all the blog data. This leads to changes in the data of the tags after selecting a tag. I need to find a method or API to always Get all the blog list data.

Do you have any suggestions for this problem? @liuxiran  @1502shivam-singh

website/src/theme/BlogSidebar/index.js
Baoyuantop: `practical case` has something special. I can’t directly set the route according to it, I need to set it as 
`practical-case`, So when I read the state from the route, I forgot to restore it. This is a bug.

liuxiran:
>I found that my problem lies in the fact that the data of the tags on the left depends on the list data on the right instead of all the blog data. This leads to changes in the data of the tags after selecting a tag. I need to find a method or API to always Get all the blog list data.

cc @bzp2010 help to take a look, thanks

liuxiran (approved):

liuxiran:
I have created a new issue to deal with the tag interaction logic, see: #554 , we need to find a better way to fix them.

bzp2010 (approved):

PR #561, opened 2021-09-07
feat: add sync error handler

bzp2010:
Changes:

It is modified that when a document synchronization error is encountered during deployment CI execution, there is no warning action.


liuxiran (approved):

liuxiran:
cc @Baoyuantop @qier222 @iamayushdas to have a look, thanks

juzhiyuan (approved):

PR #576, opened 2021-09-09
chore: bump docusaurus to 2.0beta6

bzp2010:
Changes:

Upgrade docusaurus to over version 2.0.0-beta.6


website/src/theme/DocSidebar/index.tsx
```diff
@@ -0,0 +1,152 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
```
liuxiran, line 7: Hi @bzp2010 it seems that this file has been  rewritten, would you please provide more details about this? thanks a lot

website/package.json
```diff
@@ -12,12 +12,18 @@
...
+    "@types/react": "^17.0.20",
+    "@types/react-helmet": "^6.1.2",
+    "@types/react-router-dom": "^5.1.8",
+    "babel-plugin-styled-components": "^1.12.0",
+    "typescript": "^4.4.2"
   },
   "dependencies": {
-    "@docusaurus/core": "2.0.0-beta.0",
-    "@docusaurus/preset-classic": "2.0.0-beta.0",
+    "@docusaurus/core": "2.0.0-beta.6",
```
liuxiran, line 25: could it be possible for us use `^2.0.0-beta.6` then we can use the latest docusaurus every time?

liuxiran:
Hi @bzp2010 please fix the conflict file when you have time, thanks

also cc @qier222 @Baoyuantop @iamayushdas to have a review, thanks

juzhiyuan:
cc @qier222 to have a look, thanks!

website/package.json
bzp2010: If there is a break change that will crash our CI, I recommend not using the latest version before the release version is launched.

bzp2010:
Hi @liuxiran @LiteSun @juzhiyuan,
Here I provide a full preview address. The key modification is to add the mobile version sidebar adaptation and version selection on the document page.

[link here](https://upbeat-kare-734b69.netlify.app)

website/src/theme/DocSidebar/index.tsx
bzp2010: I don't know more information. Previous versions of this file have this header text.

website/package.json
liuxiran: got you, thanks @bzp2010 , then we need to keep an eye on the developments in the docusaurus community, we can follow the updates when it has important upgrades

website/src/theme/DocSidebar/index.tsx
liuxiran: not for the header text, but for the whole file, Is it because docusaurus has upgrade the template itself?

liuxiran:
The mobile experience is better after the upgrade, and `back to top` button is also very useful.

I noticed there is a little problem on css in blog page, we can create an issue after this pr got merged if there is nothing to modified. thanks
image

Baoyuantop:
> The mobile experience is better after the upgrade, and `back to top` button is also very useful.
> 
> I noticed there is a little problem on css in blog page, we can create an issue after this pr got merged if there is nothing to modified. thanks
> image

It may be that some APIs have changed, I will pay attention to this issue

bzp2010:
> The mobile experience is better after the upgrade, and `back to top` button is also very useful.
> 
> I noticed there is a little problem on css in blog page, we can create an issue after this pr got merged if there is nothing to modified. thanks

Yes, this problem will be fixed in the following PR.

liuxiran (approved):

website/docusaurus.config.js
```diff
@@ -241,7 +241,7 @@ module.exports = {
         blog: {
           path: "blog",
           // see https://github.com/facebook/docusaurus/issues/5353
```
Baoyuantop:
```suggestion
```

bzp2010: fixed

website/src/theme/DocSidebar/index.tsx
bzp2010: The reason is that this template used JavaScript before, but now it is implemented by TypeScript. After that, the remaining templates will be migrated to TypeScript. This will provide a better coding experience.

Baoyuantop (approved):

juzhiyuan (approved):

PR #616, opened 2021-09-18
fix: revert blog page

Baoyuantop:
Changes:

Restore the blog page to the default theme.

Screenshots of the change:

image

Issue #490, opened 2021-08-20
Request help: I want to view all the blogs index in one page


liuxiran:
**Actual Behaviour**

Rensently, when I visit our [blog page](https://apisix.apache.org/blog), I found that it is hard for me to know what blogs we have, for there are only five blogs in menu list. I have to click `next` button to go to next page 

![image](https://user-images.githubusercontent.com/2561857/130215447-7f7a2f57-7418-4ec4-9317-5548e2077669.png)

I wonder if we can change our blog index page to a guided page, we can found all the blogs, no matter by name list, keyword list, and so on.

**Expected Behaviour**
Help me to found a better organized way easily view all blogs, thanks a lot

bzp2010:
In fact, more items can be displayed here. We only need to adjust them in the configuration file, but the effect is not nice.

We may need to customize the page to solve this problem.

1502shivam-singh:
> Recently, when I visit our [blog page](https://apisix.apache.org/blog), I found that it is hard for me to know what blogs we have, for there are only five blogs in menu list. I have to click `next` button to go to next page
>
> I wonder if we can change our blog index page to a guided page, we can found all the blogs, no matter by name list, keyword list, and so on.
>
> Help me to found a better organized way easily view all blogs, thanks a lot

So I did some look out and found that the most commonly used strategies are -

- Blogs thumbs come in a list from most recent to past ones, and are paginated to display certain amount on one page. Also, they provide feature to look up blogs under different sub categories
- On demand load of blogs thumbs, like the read more stories thing on google blog (check in link below)

Check out these examples for some ideas and we can discuss on the ones we like.

- https://blog.google/
- https://stripe.com/blog
- https://github.blog/
- https://www.netlify.com/blog/
- https://ionicframework.com/blog/

If we want to go complete vanilla then [React blog website](https://reactjs.org/blog/) is best bet, given that I believe React website is also based on docusaurus (but personally I think their site is really confusing to traverse among blogs, when you have tried blog sections like on our website).
EDIT: React blog is actually based on Gatsby not docusaurus.

liuxiran:
Great @1502shivam-singh , thanks for such careful research~

Among those sites listed above, I prefer https://github.blog/ 

* Categorise blog posts to make it easier for users to focus on target areas faster
* Articles are in chronological order from new to old, so that users can easily see the latest blog posts
* Topping the most popular posts in the current category is also convenient for users to access
* Turn the page to view, in line with user browsing habits
[...]

PR #641, opened 2021-10-05
feat: update the blog page

### LINKED ISSUE #470: support tag and count for our blog
## Actual Behavior

At the moment, our blog only shows the title, author and other information in the summary page, 

for example:

image

## Expected Behavior

We need to add some tags for each blog, and at the same time, viewers can search blogs via tags, so that it will be easier for viewer to get what they want

for example:

https://quickwit.io/blog

image

https://developer.axioms.io/blog

image

guoqqqi:
Changes:
image
image

1. Left sidebar replaced with tags

Screenshots of the change:

resolve https://github.com/apache/apisix-website/issues/470


guoqqqi:
image
I am not currently adding too much styling to the header page and I have a few problems that I need help with.
1. if I add a background colour or image to the header section, it doesn't seem to match the style of the other pages, like the team etc page
image

2. there seems to be some empty space and I'm torn between removing it and adding the header, but I think it looks good too, it just looks like something is missing

cc @juzhiyuan @1502shivam-singh @Baoyuantop @bzp2010 @liuxiran

website/src/theme/BlogPostItem/index.js
```diff
@@ -0,0 +1,134 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
```
juzhiyuan, line 4: This file is copied from Facebook?

guoqqqi: The file is a commit of the PR I recovered under https://github.com/apache/apisix-website/pull/616 and it appears to be a shell CLI generated file, https://docusaurus.io/docs/cli#docusaurus-swizzle-sitedir
cc @Baoyuantop have a look

website/src/theme/BlogTagsPostsPage/index.js
```diff
@@ -0,0 +1,68 @@
...
+import Layout from '@theme/Layout';
+import BlogPostItem from '@theme/BlogPostItem';
+import BlogSidebar from '@theme/BlogSidebar';
+import { ThemeClassNames } from '@docusaurus/theme-common'; // Very simple pluralization: probably good enough for now
+
+function BlogTagsPostPage(props) {
+  const { metadata, items } = props;
+  const { name: tagName } = metadata;
+  const [, setTagsCount] = useState();
+  const tagsTotal = JSON.parse(localStorage.getItem('tagsTotal'));
```
1502shivam-singh: Will pushing the definition of `tagsTotal` in useEffect cause some errors or unwanted behaviour ?

website/src/theme/BlogPostItem/index.js
1502shivam-singh: Yup, the blog page is not so customisable directly, so docusaurus has this functionality to `swizzle` out elements,
Refer to this - [`swizzle`](https://github.com/apache/apisix-website/issues/490#issuecomment-911850832)

1502shivam-singh:
> 2. there seems to be some empty space and I'm torn between removing it and adding the header, but I think it looks good too, it just looks like something is missing

Absence of balancing visual element i guess (check example below). Though, it seems trivial to me for now, we can work on it later maybe.

image

the dots graphic and slight color change does that here. We can too put something less focus wanting, but a balancing visual element in the right or left area for some balance.

Though, this one looks fine too.

juzhiyuan (approved):

bzp2010 (approved):

PR #643, opened 2021-10-06
docs: how to join slack?

juzhiyuan:
I'd like to add a post under the General Documentation, users could join Slack according to 2 ways.


website/docs/general/slack.md
```diff
@@ -0,0 +1,19 @@
+---
+id: slack
+title: Slack
```
liuxiran: how about use 'how to join slack' here?

juzhiyuan: A little long 🤔

bzp2010: Can we change it to `community` or something like it? We can provide more communication way on this page.

liuxiran: > Can we change it to community or something like it? We can provide more communication way on this page.

agree

juzhiyuan: ok

bzp2010 (approved):
LGTM

liuxiran (approved):

PR #673, opened 2021-10-25
feat: add event poster card datalization render

bzp2010:
Changes:

Remove 10.23 meetup ad on index page.

### DIFF
--- a/website/static/data/event-poster-card.json
+++ b/website/static/data/event-poster-card.json
@@ -1,0 +1,6 @@
+{
+  "show": false,
+  "expire": "2021-10-25",
+  "image": "/blog/2021/10/18/meetup",
+  "link": "https://static.apiseven.com/202108/meetup-card.png"
+}
--- a/website/static/data/event-poster-card.json
+++ b/website/static/data/event-poster-card.json
@@ -1,6 +1,6 @@
 {
   "show": false,
   "expire": "2021-10-25",
-  "image": "/blog/2021/10/18/meetup",
-  "link": "https://static.apiseven.com/202108/meetup-card.png"
+  "image": "https://static.apiseven.com/202108/meetup-card.png",
+  "link": "/blog/2021/10/18/meetup"
 }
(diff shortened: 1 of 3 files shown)


iamayushdas (approved):
Just check the failing build, otherwise its fine

website/static/data/event-poster-card.json
```diff
@@ -0,0 +1,6 @@
+{
+  "show": false,
+  "expire": "2021-10-25",
+  "image": "/blog/2021/10/18/meetup",
+  "link": "https://static.apiseven.com/202108/meetup-card.png"
```
juzhiyuan: 🤔

website/static/data/event-poster-card.json
```diff
@@ -0,0 +1,6 @@
+{
+  "show": false,
+  "expire": "2021-10-25",
```
juzhiyuan, line 3: Why do we need both show & expire?

1502shivam-singh: Same, `expire` field is understandable, but what's `show` field for ? Some sort of safeguard is it ?

website/static/data/event-poster-card.json
1502shivam-singh: I like this style, all changes needed are self explanatory and at one place and don't need codebase changes, so someone not involved much with the codebase can also do this. Not sure about the `show` field though.

website/static/data/event-poster-card.json
bzp2010: @juzhiyuan, @1502shivam-singh The `show` overwrites anything that does not exceed the expiration time but does not want to be displayed.

juzhiyuan (approved):

guoqqqi (approved):

website/static/data/event-poster-card.json
bzp2010: fixed

LiteSun (approved):

PR #708, opened 2021-11-04
fix: update sitemap.xml's loc

### LINKED ISSUE #705: Prevent from 301 redirect when visit Apache APISIX's website
**Actual Behaviour**

Hi, when we visit one URL which does not end with `/`, our website will redirect us to a new URL that ends with `/`, after searching, I find that it's because the ASF's Apache Server returns a 301 status code.

Why does this matter to us? Bcoz a 301 redirect will reduce SEO rank according to https://help.ahrefs.com/en/articles/2405181-3xx-redirect-in-sitemap-error-in-site-audit .

I have submitted a ticket here https://issues.apache.org/jira/browse/INFRA-22482 , wait for Infra's reply.

BTW, not only Apache APISIX's website is affected by this policy, Apache SkyWalking's website will return 301 too. cc @wu-sheng

juzhiyuan:
fix https://github.com/apache/apisix-website/issues/705

Changes:

After building sites and simap.xml gets generated, we use a script to add a `/` to loc.


bzp2010 (approved):
LGTM

liuxiran (approved):

PR #717, opened 2021-11-12
feat: remove Events from homepage

yzeng25:
Changes:
Since Events were migrated into Blogs with `Events` tag, the Events section on homepage becomes somewhat useless. Therefore, I would like to remove Events from homepage.

Before:
image

After:
image


juzhiyuan:
We removed it since it's maintained manually, I'm considering if we need to filter Events from all posts and generated the Events datasource? 🤔

I'm not quiet sure, cc @1502shivam-singh @bzp2010

yzeng25:
> We removed it since it's maintained manually, I'm considering if we need to filter Events from all posts and generated the Events datasource? 🤔
> 
> I'm not quiet sure, cc @1502shivam-singh @bzp2010

I would agree to keep them there if any sort of auto-generation tools can save us from manually updating it.

1502shivam-singh:
I think we should try to automate the process of adding events information, rather than removing it whole. It can act as a funnel to the recent events happening with APISIX, having a tab is nice but putting information upfront is better i believe, easier for users to see.
Let alone the fact that after removing the events list, the `Stay updated about APISIX` section does not see well in there any more.
Marking blogs that can act as events with some tag and then fetching the most recent few (or anyother automation) is a better alternative I feel.

yzeng25:
Hey guys @1502shivam-singh @juzhiyuan , I created another PR to manually update the events for now, see https://github.com/apache/apisix-website/pull/718 for details.

If you can work on the auto-generation tools or scripts, that would be great.

juzhiyuan:
That's great!! I will create a issue to track Events automatically generation :)

juzhiyuan:
So close this PR :)

PR #718, opened 2021-11-12
feat: update Events

yzeng25:
Changes:

update Events on homepage

Reference: https://apisix.apache.org/blog/tags/events
 
Before
image

After
image

### DIFF
--- a/website/static/data/events.json
+++ b/website/static/data/events.json
@@ -1,4 +1,16 @@
 [
+  {"title": "Weekly Report｜10.15-10.31 Feature Highlights Update in Progress",
+    "fileName": "2021-11-02-weekly-report-1031"
+  },
+  {"title": "Webinar | Apache APISIX × Apache SkyWalking Online Meetup",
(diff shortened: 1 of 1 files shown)


juzhiyuan (approved):

bzp2010 (approved):

website/static/data/events.json
```diff
@@ -1,4 +1,16 @@
 [
+  {"title": "Weekly Report｜10.15-10.31 Feature Highlights Update in Progress",
```
1502shivam-singh, line 2: Okay, so now as the titles have gotten longer, hierarchical difference between event title and date can't be made. So I suggest we assign `event-card-time` as different color (this maybe - #6a6a6a) and increase top (to 55% will do)

juzhiyuan: good idea, I will merge this PR first, then would you please do some changes then update it?

1502shivam-singh: Sure

PR #772, opened 2021-11-25
docs: update topic to avoid displaying issue

### LINKED ISSUE #771: Bug: display Blog(EN) page has two duplicate articles
**Actual Behaviour**

Now the Blog page has two duplicate articles about the CVE-2021-43557 announcement

**Expected Behaviour**

The page should display the two pieces of content in order：
1. https://apisix.apache.org/blog/2021/11/23/cve-2021-43557-research-report/
2. https://apisix.apache.org/blog/2021/11/23/cve-2021-43557/

**Steps to reproduce it**

 

**Screenshots of the issue**

image

**Would you like to work on the issue?**

No，need another person help to check.

yzeng25:
Fixes: #771

Changes:

Currently, the website displays two identical articles on the first page of blog. 

After some researching and testing, I find the cause of this problem. While it is completely fine to post multiple articles in the same date directory, the `title` of articles under the same directory must be identical and unique. 

In other words, when adding two or more articles in the same date directory, you need to make sure each article's `title` must be different.

This problem is caused by two articles with the same `title`. 

image

image

After updating the title of an article, then `yarn start` run it locally, it works out fine.

image

In addition, I also checked the files in `apisix-website/website/i18n/zh/docusaurus-plugin-content-blog/2021/11/23/` directory and run it locally. It works fine, so there is no need to modify files in this directory.

For future changes, if you are adding two or more articles in the same date directory, please make sure they have different `title`. Use `yarn start` to run it locally before submission is also suggested.

### DIFF
--- a/website/blog/2021/11/23/cve-2021-43557-research-report.md
+++ b/website/blog/2021/11/23/cve-2021-43557-research-report.md
@@ -1,5 +1,5 @@
 ---
-title: "Apache APISIX Path traversal in request_uri variable(CVE-2021-43557)"
+title: "Apache APISIX request_uri variable is not properly controlled, with risk of path penetration"
 author: "Marcin Niemiec"
 authorURL: "https://github.com/xvnpw"
 authorImageURL: "https://avatars.githubusercontent.com/u/17719543?v=4"


SylviaBABY (approved):

juzhiyuan (approved):

juzhiyuan:
Good catch!! This may be a bug from Docusaurus?

yzeng25:
> Good catch!! This may be a bug from Docusaurus?

Well, not necessarily. 

It is most people's first instinct to look at file names when the above problem occurs. 

File name is used to generate urls, thus duplicated file names is prohibited from the very beginning, i.e. creating two files with the same file names in the same directory is not allowed.

But if you take a step back and have a look: in what kind of scenario would a person create two or more articles with the same `title`? Basically none except done by accident.

With this being said, each article has a unique title is a very common-sense thing. I would rather call this a human-like thinking logic than a bug.

juzhiyuan:
>  I would rather call this a human-like thinking logic than a bug.

😄

bzp2010 (approved):

iamayushdas (approved):

liuxiran (approved):

LiteSun (approved):

PR #884, opened 2022-02-17
feat: support language detection

### LINKED ISSUE #743: support language detection
**Actual Behaviour**

When users visit apisix.apache.org, our website will serve content in English by default.

**Expected Behaviour**

For users in Chinese, we could provide default Chinese contents for a quick and better involving experiment.

**Solution**

According to https://docusaurus.io/docs/i18n/introduction#i18n-non-goals, Docusaurus will not do language detection, see https://github.com/facebook/docusaurus/issues/5838#issuecomment-955278743 .

Maybe we could use this package? https://www.npmjs.com/package/detect-browser-language

oil-oil:
Fixes: #743

Changes:

Added a method to check the browser language and jump to the Chinese page if the language is Chinese.

Screenshots of the change:

no

### DIFF
--- a/website/detectLang.js
+++ b/website/detectLang.js
@@ -1,9 +1,11 @@
-const browserLang =
-  (navigator.languages && navigator.languages[0]) ||
-  navigator.language ||
-  navigator.userLanguage;
-const currentLang = window.location.pathname.substring(1, 3);
-if (browserLang === "zh-CN" && currentLang !== "zh") {
-  const url = window.location.origin + "/zh" + window.location.pathname;
-  window.location.replace(url);
+if (typeof window !== "undefined") {
+  const browserLang =
+    (window.navigator.languages && window.navigator.languages[0]) ||
+    window.navigator.language ||
+    window.navigator.userLanguage;
+  const currentLang = window.location.pathname.substring(1, 3);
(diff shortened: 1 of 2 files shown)


bzp2010:
One problem I found now is that when a user using Chinese needs to view an English document, it will be brought back to Chinese mode again and we need to allow users to manually switch to English documents.

oil-oil:
> One problem I found now is that when a user using Chinese needs to view an English document, it will be brought back to Chinese mode again and we need to allow users to manually switch to English documents.

OK, i'll check it out.

oil-oil:
> One problem I found now is that when a user using Chinese needs to view an English document, it will be brought back to Chinese mode again and we need to allow users to manually switch to English documents.

I thought about it, but there seems to be no good solution to this problem at present. And if the Chinese content is missing in the future, it is not advisable to directly jump to the Chinese page.🧐

juzhiyuan:
> 

Yes, we could have more discussions on the issue before coding 🤗 I will add discuss label to that issue.

yzeng25 (approved):

SkyeYoung:
This maybe should be closed.

juzhiyuan:
Indeed. We have implemented :)

PR #907, opened 2022-02-26
fix: hide color mode switch in mobile

### LINKED ISSUE #872: UI problem of switching color mode in mobile browser
**Actual Behaviour**

The switch keep show dark mode, when reset to light mode

**Expected Behaviour**

Actually, color mode switch should not show (As shown in the view on the large screen). But if we need to keep this switch in the phone, we need to refresh the state of the switch after resetting back to light mode.

**Steps to reproduce it**

As shown in GIF below

**Screenshots of the issue**

**Would you like to work on the issue?**

Yes, but I don't have time now.

SkyeYoung:
Fixes: #872

Changes:

- Split the code of hiding color mode switch to another `useEffect`
- Use `useWindowType` to check if it is mobile
- Hide color mode switch in mobile 

Screenshots of the change:

none

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -1,6 +1,7 @@
 import React, { useState, useEffect, useLayoutEffect } from "react";
 import useThemeContext from '@theme/hooks/useThemeContext';
 import Layout from "@theme/Layout";
+import useWindowType from "@theme/hooks/useWindowSize";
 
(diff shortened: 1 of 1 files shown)


1502shivam-singh (approved):
Nice. appreciate this subtle improvement :)

yzeng25 (approved):

PR #937, opened 2022-03-03
docs: update contents for SEO

yzeng25:
Changes:

- Rename files in a more semantic way
- Add Apache APISIX intro to articles


SylviaBABY (approved):

guitu168 (commented):
Do we need to update keywords?

yzeng25:
> Do we need to update keywords?

Nope, they should be fine.

moonming (changes requested):
1. please add `API Gateway` in URI.
2. add `API gateway` in keywords.

moonming (changes requested):
There are still many articles that have not been updated

PR #958, opened 2022-03-11
feat: improve contribute page's seo

SkyeYoung:
Changes:

- improve contribute page's seo

Screenshots of the change:

none


bzp2010 (approved):

juzhiyuan (approved):
Do we need to set up a Header for each page? If Docusaurus supports reuse, that would be much better.

SkyeYoung:
> Do we need to set up a Header for each page? If Docusaurus supports reuse, that would be much better.

I think it should support it, I will check it.

PR #978, opened 2022-03-29
docs: test single lang zh-CN

yzeng25:
Changes:

For test purposes only.


yzeng25:
Close as the demo and test purpose are met.

PR #1062, opened 2022-04-28
feat: support Ahrefs verification

juzhiyuan:
**Changes:**

Added the meta tag for verification


guoqqqi (approved):

juzhiyuan:
We could only trigger `pull docs` when we change docs to save CI. 😉

yzeng25:
How to verify if it works or not?

yzeng25 (approved):

juzhiyuan:
> How to verify if it works or not?

Once it's published, then Ahrefs will pass the test directly. (By checking if HTML head has the content)

PR #1095, opened 2022-05-14
feat: support matomo from the ASF

juzhiyuan:
**Changes:**

The ASF asks to use Matomo instead of GA after July 2022, so we need to record data as early as possible by integrating this service. Visit https://analytics.apache.org/ for more information.

For PMC members, please search and read email "New data privacy policy - may impact your websites".

### DIFF
--- a/website/config/ssrTemplate.js
+++ b/website/config/ssrTemplate.js
@@ -19,6 +19,25 @@ module.exports = {
       <% it.scripts.forEach((script) => { %>
         <link rel="preload" href="<%= it.baseUrl %><%= script %>" as="script">
       <% }); %>
+      <!-- Matomo from the ASF -->
+      <script>
+        var _paq = window._paq = window._paq || [];
+        /* tracker methods like "setCustomDimension" should be called before
+      "trackPageView" */
+        /* We explicitly disable cookie tracking to avoid privacy issues */
+        _paq.push(['disableCookies']);
+        _paq.push(['trackPageView']);
+        _paq.push(['enableLinkTracking']);
+        (function() {
+          var u="https://analytics.apache.org/";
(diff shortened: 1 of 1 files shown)


MartijnVisser (approved):
From a Matomo implementation perspective, this is looking good. 
I also did a quick check for both Google Tag Manager / Google Analytics code in this repo, but couldn't find it. I did see that the APISIX website loads them, but I guess they're coming from somewhere else? Don't forget to remove those :)

Edit: found the GTM implementation, which probably loads the Google Analytics one. See https://github.com/apache/apisix-website/blob/master/website/docusaurus.config.js#L433-L435

moonming (changes requested):
Are these statistics public? Will this raise bigger privacy concerns?

MartijnVisser:
> Are these statistics public? Will this raise bigger privacy concerns?

Yes, statistics are public. The entire point why we set this up as ASF is to minimize privacy impact (that's why we don't track anything cookie/identifier based) while maintaining the data inside ASF (limited data exposure to third parties). You can have a look for yourself at https://analytics.apache.org

juzhiyuan:
> Are these statistics public? Will this raise bigger privacy concerns?

@moonming, I have explained why and how in this PR's description, as @MartijnVisser says, you could visit the ASF's Analytics service.

I had to use the new service early because we would lose past historical data metrics after switching to the new statistics service. As for the historical service, the removal will be completed before the time required by the Foundation. cc @MartijnVisser

yzeng25 (approved):

SkyeYoung (approved):

juzhiyuan:
ping @moonming

bzp2010 (approved):

SkyeYoung:
Let's merge and observe?

moonming (approved):

PR #1092, opened 2022-05-12
fix: remove error pages

SkyeYoung:
Fixes: #[Add issue number here]

Changes:

some components should not be rendered as pages

Screenshots of the change:

PR #1109, opened 2022-05-21
chore: add jest. add project vscode recommendation. 

barchielAbyss:
## Styles: 
@SkyeYoung 
abort :
> commit HASH `a2b86a5f3cbb6482118d6ea9543fbad56963f1b7`
> commit PR [fix: remove error page](https://github.com/apache/apisix-website/pull/1092)

filePath: `apisix-website/website/src/components/HeroCanvas.tsx`
remove raw-loader package, change 2 `es module` export Object

```ts
interface HeroCanvas {
    fragment: string
    vertex: string
}
```

## Changes:
1. added the test framework `jest`, and git push auto check
2. added project VsCode recommendation config

## WantChange:
@SkyeYoung I think use `shell` run githook is not good, can i write in `package.json` like this?

```json
// package.json
...
"script": {
   "lint:eslint": "eslint --fix"
},
 "husky": {
    "hooks": {
      "pre-commit": "lint-staged",
      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
    }
  },
...
```

```
// lint-staged.config

module.exports = {
  '*.{js,ts}': ['yarn lint:eslint', 'yarn test'],
}
```


SkyeYoung:
Sorry, I need to point out some problems.

Before you do something, please consider as much as possible whether we really need these improvements, and if you don't know, please create a proposal.

For example, do we really need a configuration for vscode? Does your configuration work for everyone? I think the answer is obviously no.

For example, does the addition of jest make sense? Does it cover all use cases that need to be tested?

There are other problems that I don't want to enumerate.

Also, I recommend doing one thing per PR, not everything.

SkyeYoung:
> I think use shell run githook is not good, can i write in package.json like this?

All transactions will eventually run as a shell after parsing.

If I remember correctly, so why is it _not good_? We don't have to rewrite everything in JS.

SkyeYoung:
> remove raw-loader package, change 2 es module export Object

I also want to remove rawloader, but I don't think it should be done this way.

We should configure webpack5 directly, this is a capability that docusaurus has not provided so far.

I'm making some efforts to make our dependency on docusaurus as little as possible. But it's a step-by-step process, not removing configurations that we can't handle at the moment

SkyeYoung:
I know you put a lot of effort and time into this PR, but I feel like I should point out your mistakes earlier.

juzhiyuan:
Hi @barchielAbyss, as @SkyeYoung said:

1. We should submit proposals to explain why we need this new feature and how to implement it before submitting PRs, or this project will lose control and become a hodgepodge.
2. If most people like your proposal and say "Yes", you could start contributing codes, but you need one PR to do one Thing.

> added the test framework jest

I'm not clear why we need Jest to test pages 😂 This is more like a "Marketing" project than an "Application" project.

juzhiyuan:
> For example, do we really need a configuration for vscode? Does your configuration work for everyone? I think the answer is obviously no.

For VSCode, I tend to maintain a VSCode configuration file, as I use VSCode, and I know some active contributors also use VSCode. It will not work for everyone but work for most :)

barchielAbyss:
> > remove raw-loader package, change 2 es module export Object
> 
> I also want to remove rawloader, but I don't think it should be done this way.
> 
> We should configure webpack5 directly, this is a capability that docusaurus has not provided so far.
> 
> I'm making some efforts to make our dependency on docusaurus as little as possible. But it's a step-by-step process, not removing configurations that we can't handle at the moment

emm, docusaurus can configure it,  maybe u can see [there](https://github.com/algolia/docsearch-website/blob/master/plugins/my-loaders/index.js),

and `docusaurus.config.js` add
```
plugins: ['your-plugins']
```

seted webpack5 after,  will corrects typescript checks,

I try 2 add, but it is not work
```
// typings/index.d.ts

declare module '*.txt' {
  const content: string;
  export default content;
}

declare module '*.vert' {
  const content: string;
  export default content;
}

declare module '*.frag' {
  const content: string;
  export default content;
}
```

barchielAbyss:
> > I think use shell run githook is not good, can i write in package.json like this?
> 
> All transactions will eventually run as a shell after parsing.
> 
> If I remember correctly, so why is it _not good_? We don't have to rewrite everything in JS.

We may need many stages of checks, such as `pre-commit`, `commit-msg`, if different files are required to maintain it with shell commands, the check stage is likely to be added according to certain conditions, so that the new file There will be more and more. If it is your design idea to maintain the stage through files, I think it is feasible, but I recommend only changing the rules of the content of the checked files, rather than adding new files to add process configuration.

barchielAbyss:
> Sorry, I need to point out some problems.
> 
> Before you do something, please consider as much as possible whether we really need these improvements, and if you don't know, please create a proposal.
> 
> For example, do we really need a configuration for vscode? Does your configuration work for everyone? I think the answer is obviously no.
> 
> For example, does the addition of jest make sense? Does it cover all use cases that need to be tested?
> 
> There are other problems that I don't want to enumerate.
> 
> Also, I recommend doing one thing per PR, not everything.

I understand your concern about why the `vscode` folder needs to be uploaded. It is because everyone may modify the code of this project, so you should provide a basic configuration of the vscode for this project, which may increase Your code review work, see if anyone has modified this file, but the advantage of this is that everyone will have a default configuration in the project's workspace, which can greatly reduce their configuration of this project time

barchielAbyss:
> Hi @barchielAbyss, as @SkyeYoung said:
> 
> 1. We should submit proposals to explain why we need this new feature and how to implement it before submitting PRs, or this project will lose control and become a hodgepodge.
> 2. If most people like your proposal and say "Yes", you could start contributing codes, but you need one PR to do one Thing.
> 
> > added the test framework jest
> 
> I'm not clear why we need Jest to test pages 😂 This is more like a "Marketing" project than an "Application" project.

When I write any script, I have the habit of writing test cases first, if you think it is not applicable in this project, I can delete it first

barchielAbyss:
> > > I think use shell run githook is not good, can i write in package.json like this?
> > 
> > 
> > All transactions will eventually run as a shell after parsing.
> > If I remember correctly, so why is it _not good_? We don't have to rewrite everything in JS.
> 
> We may need many stages of checks, such as `pre-commit`, `commit-msg`, if different files are required to maintain it with shell commands, the check stage is likely to be added according to certain conditions, so that the new file There will be more and more. If it is your design idea to maintain the stage through files, I think it is feasible, but I recommend only changing the rules of the content of the checked files, rather than adding new files to add process configuration.

If you care about file matching rules, you can configure them like this:

```
// lint-staged.config

module.exports = {
  '*': ['yarn lint:eslint'],
  '*.{js,ts}': ['yarn test'],
}
```

SkyeYoung:
Let's divide these discussions into several issues:

1. #1110
2. #1111
3. #1112

barchielAbyss:
> [Site][Discussion]: remove raw-loader #1111

I like the way you work

PR #995, opened 2022-03-31
fix: Explicitly set the rel=canonical of the OpenTelemetry blog post

nfrankel:
Duplicate content is frowned upon by Google. This blog post has been already published on the OpenTelemetry web site. To avoid being flagged as duplicate and incur a SEO penalty, we need to reference the "original".

Alternatively, we could ask the OpenTelemetry team to point to our post. But it will probably take more time and it's not worth it IMHO.

PR #1126, opened 2022-05-31
docs: add canonical links for cross-posted documents

### LINKED ISSUE #999: [Docs]: Track canonical_url's actual behavior
### Current State

**Description of the current state/issue**:

Some of the APISIX blogs are cross-posted on other platforms such as Medium or other communities' blog websites. While Medium provides canonical url to point to the original post, other communities' blog websites lack of this feature. This could create negative effects on SEO.

Currently, we are unsure about the actual effect of canonical_url, and we are going to add canonical_url to a recently published blog to see if it brings difference, see https://github.com/apache/apisix-website/pull/995. 

**Link to the page**:
https://apisix.apache.org/blog/2022/02/28/apisix-integration-opentelemetry-plugin/

**Screenshots (if relevant)**:
none.

### Desired State

**Description of the change you are proposing**:

We should review GA stats after a month or so, preferably early May, to see if canonical_url actually works. If it works, then we should do it for each cross-posted blog that is posted on platforms other than Medium. If it does not work, we should rollback the change in this pull request.

**How to determine if it has effect**

**Expected Changes**

According to the definition of usages of canonical_url, it will reduce duplicate and return only one result. In this case, Medium applies canonical_url and points to APISIX's blog, then we apply canonical_url on APISIX's blog and points to OpenTelemetry's blog.

In the end, Medium's link won't show up, ideally APISIX's blog won't show up either, and only OpenTelemetry's blog will show up(not sure if nested canonical_url works like this).

**View Count**: if its view count of APISIX's blog conentent does not differ much from other blogs posted in the same period of time, then we can say canonical_url  has less effect, we should reconsider adding it to all cross-posted articles.

**Search Engine Results**: 
Before adding canonical_url, when typing  "apisix opentelemetry" or "opentelemetry apisix" in Google, the results are:
image
image

We'd expect only OpenTelemetry's blog will show up, or at least both APISIX's blog and OpenTelemetry's blog will show up, with OpenTelemetry's blog in front of APISIX's blog. Otherwise, this pull request is proven to be wrong.

### Would you like to fix this issue?

Yes! Count me in

yzeng25:
Fixes: #999

Changes:
1. Add canonical link for Keycloak blog, zh-cn and en-us.
2. Add canonical link for Dapr blog, zh-cn and en-us.
3. Add canonical link for OpenTelemetry blog, zh-cn only, since en-us were already done in https://github.com/apache/apisix-website/pull/995/

### DIFF
--- a/website/blog/2021/11/17/dapr-with-apisix.md
+++ b/website/blog/2021/11/17/dapr-with-apisix.md
@@ -17,6 +17,10 @@ tags: [Technology]
 
 <!--truncate-->
 
+<head>
+    <link rel="canonical" href="https://blog.dapr.io/posts/2022/01/13/enable-dapr-with-apache-apisix-ingress-controller/" />
+</head>
+
 Essentially, the Apache APISIX controller will configure the same standard DAPR annotations to inject DAPRD sidecar. Exposing this sidecar allows external applications to communicate with applications in the cluster that have Dapr enabled.
 
 The following diagram shows the architectural flow of the actual project:
--- a/website/blog/2021/12/10/integrate-keycloak-auth-in-apisix.md
+++ b/website/blog/2021/12/10/integrate-keycloak-auth-in-apisix.md
@@ -22,6 +22,10 @@ tags: [Technology,Authentication]
 
 <!--truncate-->
 
+<head>
+    <link rel="canonical" href="https://www.keycloak.org/2021/12/apisix" />
+</head>
+
 ![Keycloak-APISIX Integration](https://static.apiseven.com/202108/1639129658486-393e8a3a-ccf2-496d-9b46-4db741bd6e55.png)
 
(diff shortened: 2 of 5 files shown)


yzeng25:
@juzhiyuan @SkyeYoung @nfrankel @SylviaBABY @hf400159 please take a look when available, thanks.

nfrankel:
@yzeng25 The intent is great, but I don't know if the generation engine takes the `<head>` into account. If yes, then perfect!

SylviaBABY (approved):
LGTM

guitu168 (approved):

yzeng25:
> @yzeng25 The intent is great, but I don't know if the generation engine takes the `<head>` into account. If yes, then perfect!

I just checked.

#995 did this and here is the preview.

截屏2022-05-31 10 31 13

This pull request did similar things and they show up in preview environment. Here are a couple of examples.


截屏2022-05-31 10 34 06

截屏2022-05-31 10 33 19

nfrankel:
Perfect!!!

juzhiyuan (approved):

PR #1148, opened 2022-06-11
feat: remove canvas & add schema

juzhiyuan:
Changes:

- [x] added structured schema (global)
- [x] remove Canvas to improve visit performance

cc @SkyeYoung to have a review when you're available


SkyeYoung (changes requested):
image

Looks like it needs some improvement.

1502shivam-singh:
> With Canvas -

image

> Without Canvas -

image

Not a performance boost tbh (even if lighthouse might not be the most reliable). Also, the return from this big change isn't the most beneficial.
About performance issues, just to mention here, I had created a system to stop rendering on both canvases when they were not visible on the browser viewport, to squeeze out the most performance boost. 
So, I don't recommend this change to take place, as it doesn't provide a huge chunk of performance boost anyway and does more harm than good to the overall visual style and branding of the website.

juzhiyuan:
Hi @1502shivam-singh @SkyeYoung, the reason to remove the Canvas depends on two reasons:

1. Visitors report that the two canvas sections don't have a special meaning. I don't have better ideas, so I removed them in this PR.
2. https://pagespeed.web.dev/ reports that the landing page performance is terrible. It takes a long time to interact. But @1502shivam-singh is correct, as the following screenshot shows, the current site performance is even better.

Do you have ideas to improve the performance?

image

1502shivam-singh:
> 1. Visitors report that the two canvas sections don't have a special meaning. I don't have better ideas, so I removed them in this PR.

I can work on improving the content displayed on the canvas in the "Build API Gateway together" section (It was the only rushed element in the redesign) I can create a custom motion graphic to replace it maybe, and add it as a lottie. But this would take some time, need to keep it till then 😅 Or if anyone else could contribute in this regard, most welcome.

> Do you have ideas to improve the performance?

So I went back and tried to analyze the performance pre-redesign through web archive. Seems the pre-redesign one didn't exactly have the best performance either. So, we should think in this regard and then work our way through.

Compare

As of the performance decrement in this comparison (variable on multiple trials, but the numbers were fairly close on each test iteration), my best guess would be increased DOM length. But that doesn't justify for a huge decrement in performance.

juzhiyuan:
Hi @1502shivam-singh, thanks for your explanation, and I agree. 

> Or if anyone else could contribute in this regard, most welcome.

I know @SkyeYoung is also good at design style 😄 (but he is busy fixing some issues on the website); you two can communicate 😊 and I will close this PR first.

PR #1361, opened 2022-10-15
docs: update the event page with new events

Boburmirzo:

Changes:

Updated events page with new upcoming events

Screenshots of the change:

### DIFF
--- a/website/docs/general/events.md
+++ b/website/docs/general/events.md
@@ -18,103 +18,33 @@ that have a specific focus on Apache APISIX.
 
 ### Conferences
 
-#### [Code Palousa](https://www.codepalousa.com/)
-
-* **Topic** Evolving your APIs
-* **Speaker** [Nicolas Fränkel](https://github.com/nfrankel)
-* **Dates** from 17-August-2022 to 19-August-2022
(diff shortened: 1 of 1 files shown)


SylviaBABY (dismissed):

pottekkat (commented):
Will it make sense to move the "Meetups" section to the top?

Because there will always be a lot of conference talks ahead and we will only have one upcoming meetup at a time (maximum two). And if it is on top, it might be easy for people to find it.

Since workshop section is empty, should we have some text saying something like "more workshops will be announced soon".

Boburmirzo:
> Will it make sense to move the "Meetups" section to the top?
> 
> Because there will always be a lot of conference talks ahead and we will only have one upcoming meetup at a time (maximum two). And if it is on top, it might be easy for people to find it.
> 
> Since workshop section is empty, should we have some text saying something like "more workshops will be announced soon".

Sounds reasonable to me! Thanks. 

I updated it respectively.

pottekkat (approved):

guitu168 (approved):

tokers (approved):

Issue #980, opened 2022-03-29
[Site]: support structured schema (JSON-LD)


juzhiyuan:
### Current State

**Description of the current state/issue**: For each page under the Apache APISIX Website, we don't provide a structured schema for search engines, so the search engine has to try to understand each page's content meaning. Structured Schema is an Advanced SEO solution, please refer to the following links for more information.

![image](https://user-images.githubusercontent.com/2106987/160530278-cc064c3e-9a4d-4eaa-b23f-331b710d1caf.png)

1. https://moz.com/blog/json-ld-for-beginners
2. https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data
3. https://stackoverflow.com/questions/38670851/whats-a-script-type-application-ldjsonjsonobj-script-in-a-head-sec/38745339#38745339
4. https://support.google.com/knowledgepanel/answer/9163198?hl=zh-Hans&ref_topic=9164489
5. https://developers.google.com/search/docs/advanced/guidelines/get-started

### Desired State

Validator Tools: https://validator.schema.org/

SkyeYoung:
If we need a good type prompt, we can use https://www.npmjs.com/package/schema-dts.

nfrankel:
I'm interested to contribute on this one. It can be a quick win.

juzhiyuan:
Hi, Nicolas, Thanks!!

nfrankel:
I've checked in more detail and it's not as quick a win as I thought.

The number of items is pretty limited. I believe we can use [`How-To`](https://developers.google.com/search/docs/advanced/structured-data/how-to) and [`Article`](https://developers.google.com/search/docs/advanced/structured-data/article) quite easily. Others, such as `Learning Video` and `Software App` are a bit harder to push in.

Hence, I think I could start with one article from the blog and see what I can make out of it.

pottekkat:
@juzhiyuan @nfrankel @SkyeYoung we have to work on this to improve discoverability in Google. From experience, adding schema to the site can help in getting a "Knowledge Panel" in Google search. @SkyeYoung With your help, I can probably look into what we can do and take this forward.

nfrankel:
Yes, I wanted to work on that months ago... and did nothing. Feel free to own it

PR #1493, opened 2023-02-03
feat: add structured data for SEO

pottekkat:
Signed-off-by: Navendu Pottekkat <navendu@apache.org>

Related to #980

Adds structured data for the homepage. This is an initial setup and in the future we can automatically generate this data on every page in the site.

Screenshot 2023-02-03 at 11 16 30 AM

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -44,6 +44,33 @@ const ThemeResetComponent = () => {
   return (null);
 };
 
+// structured data for improved SEO
+const structuredData = {
+  "@context": "https://schema.org",
+  "@type": "softwareApplication",
+  name: "Apache APISIX",
+  alternateName: "APISIX",
+  description: "Open source and cloud native API gateway, based on the Nginx library and etcd.",
+  abstract: "High-performance, cloud-native, open source API Gateway from the Apache Software Foundation.",
+  url: "https://apisix.apache.org",
+  thumbnailUrl:
+    "https://raw.githubusercontent.com/apache/apisix/master/logos/apisix-white-bg.jpg",
+  sameAs: [
+    "https://twitter.com/ApacheAPISIX",
+    "https://github.com/apache/apisix",
+    "https://www.youtube.com/@apacheapisix",
+    "https://www.linkedin.com/company/apache-apisix/",
+  ],
+  isPartOf: "Apache Software Foundation",
+  maintainer: "Apache Software Foundation",
+  contributor: [
+    "API7.ai",
+    "Tencent",
+    "Zhihu",
+    "Alibaba Cloud",
+  ],
+};
+
 const Index: FC = () => (
   <Layout>
     <ThemeResetComponent />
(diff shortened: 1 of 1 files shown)


pottekkat:
Also requesting a review from @nfrankel

pottekkat:
@SkyeYoung Please let me know if I need to move the schema definition to some place else.

website/src/pages/index.tsx
```diff
@@ -44,6 +44,33 @@ const ThemeResetComponent = () => {
...
+    "https://raw.githubusercontent.com/apache/apisix/master/logos/apisix-white-bg.jpg",
+  sameAs: [
+    "https://twitter.com/ApacheAPISIX",
+    "https://github.com/apache/apisix",
+    "https://www.youtube.com/@apacheapisix",
+    "https://www.linkedin.com/company/apache-apisix/",
+  ],
+  isPartOf: "Apache Software Foundation",
+  maintainer: "Apache Software Foundation",
+  contributor: [
```
GeekEmily: What are the criteria to pick the contributors?

SkyeYoung:
> @SkyeYoung Please let me know if I need to move the schema definition to some place else.

No need I think. 🤔 But we need to improve some of the content.

website/src/pages/index.tsx
pottekkat: I just went to orbit and found top organizations that contribute to APISIX. Should we do it?

nfrankel: You need to have approval of some representative of the organizations for that.

pottekkat: Removed the section.

pottekkat:
> No need I think. 🤔 But we need to improve some of the content.

Could you share what we can improve?

SkyeYoung:
> Could you share what we can improve?

Sorry, I'm late. My thoughts are similar to what keer said, and I think it's good now that it's been revised.

juzhiyuan (approved):
Just to confirm: this only adds Schema to the homepage?

pottekkat:
> Just to confirm: this only adds Schema to the homepage?

Yes. It only does that. But AFAIK it will help with our goal to get a knowledge panel. In future, we can generate schema automatically for each page from the metadata. I have to spend more time looking into how I can achieve this or someone else can take it up.

website/src/pages/index.tsx
```diff
@@ -64,6 +85,11 @@ const Index: FC = () => (
         name="og:description"
         content="Apache APISIX is a dynamic, real-time, high-performance Cloud-Native API gateway, based on the Nginx library and etcd."
       />
+
+      <script type="application/ld+json">
```
nfrankel, line 89: Why don't you put the JSON directly here?

nfrankel:
[Google Structured Data Testing Tool](https://search.google.com/test/rich-results) isn't happy about some missing fields:

Screenshot 2023-02-06 at 09 29 42

website/src/pages/index.tsx
pottekkat: It did not work when I directly added the JSON. I looked at an example implementation, and they used a similar approach.

nfrankel:
I seem to remember Google complaining about missing field why `Blog` was easier to start with (cf. https://github.com/apache/apisix-website/issues/980#issuecomment-1156033243).

pottekkat:
@nfrankel Should we add `applicationCategory` and `operatingSystem`? Another alternative is to use https://schema.org/WebSite

Which way should we go?

nfrankel:
@navendu-pottekkat As I mentioned, it should be easier to start with `Blog`

PR #1654, opened 2023-07-25
blog: request validation

pottekkat:
Signed-off-by: Navendu Pottekkat <navendu@apache.org>

New blog post on using the request-validation plugin.

**Please don't merge this PR after approval. I will merge it when we are ready to promote it.**

### DIFF
--- a/blog/en/blog/2023/07/28/request-validation.md
+++ b/blog/en/blog/2023/07/28/request-validation.md
@@ -1,0 +1,410 @@
+---
+title: Your API Requests Should Be Validated
+authors:
+  - name: "Navendu Pottekkat"
+    title: "Author"
+    url: "https://github.com/navendu-pottekkat"
+    image_url: "https://avatars.githubusercontent.com/u/49474499"
+keywords:
+  - Request validation
+  - Best practice
+  - JSON schema
+description: A tutorial on using the APISIX request-validation plugin and why you might need to use it.
+tags: [Plugins]
+image: https://static.apiseven.com/uploads/2023/07/25/hB8seVHP_request-validation.png
+---
+
+> Adding a new layer of validation within your API gateway can be a useful design practice for a myriad of reasons. This article explores how you can configure this in Apache APISIX.
+
+<!--truncate-->
+
+<head>
+    <link rel="canonical" href="https://navendu.me/posts/request-validation/" />
+</head>
+
+One of the most overlooked aspects while designing APIs is request validation.
+
+An API gateway like [Apache APISIX](https://apisix.apache.org/), predominantly used for fine-grained traffic control, can also validate your API requests. This new layer of validation within the API gateway, along with client-side and application-side validations, can be a useful design practice for many reasons, like:
+
+1. Preventing invalid requests from being sent to backend services, reducing unnecessary load and thereby optimizing performance and reducing cost.
+2. Allowing application developers to focus solely on application/business-specific validations in their services.
+3. Adding another layer of security in front of your backend services, fending off malicious requests from ever reaching your services.
+
+In this article, you will learn how you can configure APISIX to validate requests with JSON schemas before forwarding them to your upstream services.
+
+## Creating the Schema
+
+For this example, our request will contain a purchase request sent from a client application to the payment service through the APISIX API gateway. A valid request body will look like this:
+
+```json
+{
+  "uid": 2374,
+  "item": "pin",
+  "quantity": 5,
+  "price": 5.0
+}
+```
+
+Each key in this request represents the following:
+
+- `uid`: Unique ID of the user. It will be empty for a guest user.
+- `item`: Item being purchased. It can be one of "sticker," "t-shirt," or "pin."
+- `quantity`: Quantity of the item being purchased.
+- `price`: Price of the item being purchased.
+
+There are a lot of tools available, like the one from [JSON formatter](https://jsonformatter.org/json-to-jsonschema) that converts JSON to JSON schema. But we will write our own schema to be more accurate:
+
+```json
+{
+  "type": "object",
+  "required": ["quantity", "price", "item"],
+  "properties": {
+    "uid": {
+      "type": "integer"
+    },
+    "item": {
+      "type": "string",
+      "enum": ["sticker", "t-shirt", "pin"]
+    },
+    "quantity": {
+      "type": "integer",
+      "minimum": 1,
+      "maximum": 50
+    },
+    "price": {
+      "type": "number",
+      "minimum": 1.0,
+      "maximum": 50.0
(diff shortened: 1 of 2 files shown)


Boburmirzo:
@navendu-pottekkat Don't we want to publish it on the API7 website?

pottekkat:
@Boburmirzo I think it is better here because it is very APISIX specific. I usually publish articles which have a more broader audience in the other website. Also it is good to keep publishing here IMO. Any strong reasons why we should publish on the other website?

Boburmirzo:
> @Boburmirzo I think it is better here because it is very APISIX specific. I usually publish articles which have a more broader audience in the other website. Also it is good to keep publishing here IMO. Any strong reasons why we should publish on the other website?

That makes sense! No, I do not have a strong reason. I remember the Growth team was pushing the idea of publishing a post first to the API7 blog. That's why I asked:)

Boburmirzo:
@navendu-pottekkat The post is nice and quite easy to understand. I love that! Would you like to make a tutorial from it later on too?

Just I am thinking it will be quite useful content for the Tutorials section on the APISIX website:

https://apisix.apache.org/docs/apisix/tutorials/expose-api/

Boburmirzo:
> @navendu-pottekkat The post is nice and quite easy to understand. I love that! Would you like to make a tutorial from it later on too?
> 
> Just I am thinking it will be quite useful content for the Tutorials section on the APISIX website:
> 
> https://apisix.apache.org/docs/apisix/tutorials/expose-api/

I can make a video tutorial using your blog post if you allow me.

pottekkat:
@Boburmirzo Yes to all. Please feel free.

pottekkat:
I can modify it to fit into the tutorials section.

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
+---
+title: Your API Requests Should Be Validated
+authors:
+  - name: "Navendu Pottekkat"
```
nfrankel: Nitpick: why the unnecessary double quotes?

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
...
+tags: [Plugins]
+image: https://static.apiseven.com/uploads/2023/07/25/hB8seVHP_request-validation.png
+---
+
+> Adding a new layer of validation within your API gateway can be a useful design practice for a myriad of reasons. This article explores how you can configure this in Apache APISIX.
+
+<!--truncate-->
+
+<head>
+    <link rel="canonical" href="https://navendu.me/posts/request-validation/" />
```
nfrankel: Careful if the post is not available when Google indexes the site

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
...
+                - t-shirt
+                - pin
+            quantity:
+              type: integer
+            price:
+              type: number
+#END
+```
+
+You can also use the Admin API for this exact configuration:
```
nfrankel: Please add that if it applies only if not in standalone-mode

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
...
+  "origin": "172.22.0.1, 223.239.0.41",
+  "url": "http://localhost/anything/anything"
+}
+```
+
+## Non-trivial Vulnerabilities
+
+Different services might use different JSON parsing implementations. Sometimes, this could lead to different parsed JSON in different services. This can open up a lot of non-trivial interoperability vulnerabilities.
+
+For example, a client can send a malicious request with duplicate keys, and two different services might end up using two different keys:
```
nfrankel: Good stuff!

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
...
+"$schema": "http://json-schema.org/draft-04/schema#"
+```
+
+## ~~Why~~ Where to Validate?
+
+It is fair to assume that you now understand why request validation might be necessary. However, you could still argue that validation can be done on the client side or within your backend services.
+
+A key point is that you might not always own your clients, or modifying client requests might be relatively easy. Adding validation in a layer you have absolute control of is the better alternative.
+
+So why can't you validate requests in your services directly? Well, after a point, it becomes too complex for the application developer to configure both request validation and application/business-specific validation. Without an API gateway layer, invalid requests still end up adding load to your backend services.
```
nfrankel: On that, I completely disagree. On the opposite, I think that's the main argument you need to make in the introduction.

Validation is business logic. You want to avoid as much business logic as possible in the gateway because of deployment synchronization issue. Putting something as trivial as body validation in the API Gateway causes sync problems for not much value - a trip from the gateway to the upstream (they should be close anyway).

Validating headers, however, is much more useful (for security) and is much less business related.

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
...
+
+Ideally, you should validate requests in all three layers with only business-specific validation in your services. The [APISIX Dashboard](https://apisix.apache.org/docs/dashboard/USER_GUIDE/) can export the configured routes to [OpenAPI format](https://spec.openapis.org/oas/latest.html), which can be used for client-side validation.
```
nfrankel: No, definitely. Especially since you may not be in charge of the client.

blog/en/blog/2023/07/28/request-validation.md
```diff
@@ -0,0 +1,410 @@
...
+```yaml title="apisix.yaml" {26-27,30-31}
+routes:
+  - uri: /anything/*
+    upstream:
+      type: roundrobin
+      nodes:
+        "httpbin:80": 1
+    plugins:
+      request-validation:
+        body_schema:
```
nfrankel: Is it possible to refer to an external file instead?

nfrankel (changes requested):

nfrankel:
@yzeng25 I'm still waiting for an answer (or a change) to my last comment, which wasn't addressed. The content is good, but the question of why you should validate in the API Gateway needs to be answered in the introduction, not the conclusion, and with convincing arguments.

As it stands, I expect pushback - and with good reasons.

blog/en/blog/2023/07/28/request-validation.md
pottekkat: It is mentioned in the introduction. A trip from the gateway to the upstream is not the issue, but sending invalid requests to the upstream unnecessarily can be avoided.

PR #1721, opened 2023-11-06
blog: Add API Versioning post

nfrankel:
Screenshot 2023-11-06 at 17 57 00

### DIFF
--- a/blog/en/blog/2023/11/09/api-versioning.md
+++ b/blog/en/blog/2023/11/09/api-versioning.md
@@ -1,0 +1,202 @@
+---
+title: API versioning
+authors:
+  - name: Nicolas Fränkel
+    title: Author
+    url: https://github.com/nfrankel
+    image_url: https://avatars.githubusercontent.com/u/752258
+keywords:
+  - DevOps
+  - Performance
+  - Web Developent
+  - Cost optimization
+  - imgproxy
+description: >
+  In my previous post Evolving your APIs, I mention the main API versioning approaches. During the talk of the same name, I sometimes get some questions on the subject. In this post, I'll detail each of them.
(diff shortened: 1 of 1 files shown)


juzhiyuan (dismissed):
The structure looks good to me. But I didn't test this thoroughly.

blog/en/blog/2023/11/09/api-versioning.md
```diff
@@ -0,0 +1,202 @@
+---
+title: Resizing images on-the-fly
```
kayx23: Why this title

nfrankel: Because I copy-pasted and forgot to update. Good catch, thanks!

kayx23 (approved):
LGTM.

Also just sharing this API versioning strategy from [Stripe](https://stripe.com/blog/api-versioning) that looks interesting:

> At Stripe, we implement versioning with rolling versions that are named with the date they’re released (for example, 2017-05-24). Although backwards-incompatible, each one contains a small set of changes that make incremental upgrades relatively easy so that integrations can stay current.

nfrankel:
Thanks for the link @kayx23. Actually, they implement the custom header-based versioning approach:

>The first time a user makes an API request, their account is automatically pinned to the most recent version available, and from then on, every API call they make is assigned that version implicitly. This approach guarantees that users don’t accidentally receive a breaking change and makes initial integration less painful by reducing the amount of necessary configuration. Users can override the version of any single request by manually setting the `Stripe-Version header`, or upgrade their account’s pinned version from Stripe’s dashboard.

yzeng25 (approved):

PR #1789, opened 2024-04-22
Fix: Added redirection to GitHub index when no hash present

### LINKED ISSUE #1783: docs: editing url valid
### Current State

in this [url](https://apisix.apache.org/edit/) and  this [url](https://apisix.apache.org/zh/edit/);
There is no response when clicking this button and it will not jump to other pages.
image

### Desired State

Automatically jump to the warehouse where modifications can be submitted

siyaramsujan:
Fixes: #1783

Changes:

Added redirection to the GitHub repository index when the path variable is empty.

### DIFF
--- a/doc/src/pages/edit.tsx
+++ b/doc/src/pages/edit.tsx
@@ -9,6 +9,8 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 const Edit: FC = () => {
   const [pathExist, setPathExist] = useState(true);
   const [isLoading, setIsLoading] = useState(false);
+  const INDEX_URL = 'https://github.com/apache/apisix';
+
   const {
     i18n: { currentLocale },
   } = useDocusaurusContext();
@@ -20,6 +22,7 @@ const Edit: FC = () => {
     const path = getPath();
 
     setIsLoading(true);
+
     fetch(path.replace('github.com', 'raw.githubusercontent.com').replace('/edit', ''))
       .then((res) => setPathExist(res.status !== 404))
       .finally(() => setIsLoading(false));
@@ -27,11 +30,17 @@ const Edit: FC = () => {
 
   const edit = useCallback(() => {
     let path = getPath();
+
     if (!pathExist) {
       const pathArr = path.replace('edit', 'new').split('/');
       pathArr[pathArr.length - 1] = `?filename=${pathArr.at(-1)}`;
       path = pathArr.join('/');
     }
+
+    if (!path?.includes(INDEX_URL)) {
+      path = INDEX_URL;
+    }
+
     window.location.replace(path);
   }, [pathExist]);
 
@@ -96,7 +105,7 @@ const Edit: FC = () => {
               <p>
                 {pathExist
                   ? 'When you are ready, click the button below to start editing the document.'
-                  : 'You can also still try to edit the document by click the below button.'}
+                  : 'You can also still try to edit the document by clicking the below button.'}
               </p>
               <button className="edit-btn" type="button" onClick={edit}>
                 Let&apos;s start editing


doc/src/pages/edit.tsx
```diff
@@ -20,18 +22,26 @@ const Edit: FC = () => {
...
   }, []);
 
   const edit = useCallback(() => {
-    let path = getPath();
+    const path = getPath();
+
     if (!pathExist) {
-      const pathArr = path.replace('edit', 'new').split('/');
-      pathArr[pathArr.length - 1] = `?filename=${pathArr.at(-1)}`;
-      path = pathArr.join('/');
```
pottekkat: Not sure if this is ever run but it seems like it is for creating a new file if doesn't exist?

Should we replace this logic?

Maybe instead we can add to this logic and set a new condition for an empty path. Because path does not exist and path is empty are different.

Again, I'm not sure if it is ever run so IDK if it would be any problem.

siyaramsujan: Hi, It looks like the code isn't creating new files. 

I think it's attempting to replace the path, for example, from '/edit/#githubrepolink' to '/new/?filename=githubrepolink' if it doesn't exist. But, we're using getPath = () => window.location.hash.slice(1); to retrieve the path, which only gives us the hash - "githubrepolink'".

So, it seems that this logic isn't being utilized. Please correct me if I'm wrong. Thanks!

pottekkat: I'm not sure, I will check with @SkyeYoung

siyaramsujan: Sure, Thank you for your review, @pottekkat.

SkyeYoung (approved):

SkyeYoung:
Thanks for the contribution! The logic should be fine.

cc @bzp2010 @guoqqqi pls review this

doc/src/pages/edit.tsx
```diff
@@ -20,18 +22,26 @@ const Edit: FC = () => {
     const path = getPath();
 
     setIsLoading(true);
+
+    if (path === '' || !path.includes('https://github.com/apache')) {
+      setPathExist(false);
+      setIsLoading(false);
+      return;
+    }
```
SkyeYoung: You can move this part of the code up 24 lines, you should not need to `setIsLoading(false)` this way.

Also `if(xxx)` can be written as `if(!path?.includes(INDEX_URL))` 🤔

doc/src/pages/edit.tsx
SkyeYoung: This logic will actually jump to the github url of "Create a new file". 

The working logic is: The document site uses English as fallback, and when documents in Chinese or other languages ​​do not exist, they will automatically be displayed in English. But at this time, `edit url` will still use the path of the corresponding language. 

For example, `/xxx` exists, but the document `/zh/xxx` does not exist. Assume that the edit url of the latter is `github/zh/xxx`. Since the document does not exist at this time, the content in `/xxx` will be displayed, but the edit url Still `github/zh/xxx`. Apparently at this point, (at least I thought so,) clicking this link should create this non-existent `/zh/xxx`.

SkyeYoung: If you follow your modified logic, it should jump directly to a non-existent path, causing github to display 404.

So I think this current logic needs to be adjusted.

SkyeYoung: Here is a practical example, please find the edit url link in the page 

https://apisix.apache.org/zh/docs/apisix/tutorials/manage-api-consumers/

SkyeYoung (changes requested):
I review it again. There are some problems. Please adjust them, Thx 😸

doc/src/pages/edit.tsx
siyaramsujan: Hi @SkyeYoung, Thanks for explaining. I finally understand how it's supposed to work, I had misunderstood before. The modified logic should redirect the user to the index URL if it doesn't exist. Should I revert it back to the logic of redirecting to GitHub create new file? Thanks again.

doc/src/pages/edit.tsx
siyaramsujan: Thanks for the review, @SkyeYoung. I'll move that code up and use `if (!path?.includes(INDEX_URL))` as you suggested. By the way, I'm currently using `INDEX_URL` as `https://github.com/apache/apisix` is that okay? Thanks!

SkyeYoung: I think it's okay.

doc/src/pages/edit.tsx
SkyeYoung: @siyaramaa Yes. Sorry for replying so late.

siyaramsujan:
Thank you, I have just made the update.

yzeng25 (approved):

yzeng25:
PTAL @SkyeYoung

SkyeYoung (approved):

PR #1820, opened 2024-08-29
fix:  the static resource loading exception encountered

guoqqqi:
When using a global proxy, the page cannot be accessed normally:

img_v3_02e7_aab53a46-04de-44b7-b61a-e264159ff01g

Changes:

1. remove unavailable CDN domains.
2. use the original domain name: apisix.apache.org


guoqqqi:
Since the CDN vendor has fixed this issue, this PR is closed.
image

PR #1832, opened 2024-11-20
chore: archive apisix documentation up to 3.8.0

bzp2010:
Changes:

Currently, each document build includes all APISIX versions from 2.13 to latest, which handles a lot of markdown documents and makes each build 40-50 minutes long. The PR archives old documents, which are fixed on the release branch, so they are not updated and do not affect reading.
I keep nearly 3 versions of documents in the build process and the old ones are moved to the archive.

We will consider extending the boundaries of the kept version if there are still requests for older documents.

### DIFF
--- a/config/apisix-versions.js
+++ b/config/apisix-versions.js
@@ -3,7 +3,7 @@
 /**
  * @type {Array<string>} version list
  */
-const versions = ['2.13', '2.14', '2.15', '3.0', '3.1', '3.2', '3.3', '3.4', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10', '3.11'];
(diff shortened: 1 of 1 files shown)


bzp2010:
**This is an improvement to the speed of CI, and older versions of the documentation are still available.**

SkyeYoung (approved):

SkyeYoung:
@bzp2010 Maybe better title: `chore: archive apisix documentation up to 3.8.0`

guoqqqi (approved):

bzp2010:
The results of this optimization were significant, with build times dropping from 40+ minutes to less than 5 minutes.

https://github.com/apache/apisix-website/actions/runs/11945981190/job/33299582500

PR #1834, opened 2024-12-02
fix: the build error **Invalid docs option "versions": unknown versions (2.15) found**

guoqqqi:

1. Because we archive apisix documentation up to 3.8.0, so we can remove the 2.15 config, it's causes the main branch build error

refer #1832 

Changes:

Screenshots of the change:


juzhiyuan (approved):

SkyeYoung (approved):
Provide a more specific title?

PR #1848, opened 2025-01-20
removed the google calendar in event calendar which is not loading

Jitmisra:
Fixes: #[removed the google calendar in event calendar which is not loading]

Changes:

Monosnap Events calendar  Apache APISIX® -- Cloud-Native API Gateway 2025-01-20 09-56-33
https://apisix.apache.org/docs/general/events/

 for now I have commented out  that link if you give me new link I will add it

### DIFF
--- a/website/docs/general/events.md
+++ b/website/docs/general/events.md
@@ -1,25 +1,15 @@
 ---
 id: events
-title: Events calendar
+title: Events
 keywords:
   - Events
   - Conferences
   - Workshops
   - Meetups
-description: This page provides information about upcoming Apache APISIX's community events where you can track meetups, conferences, and workshops.
+description: This page provides information about previous Apache APISIX's community events.
 ---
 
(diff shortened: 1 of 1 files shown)


yzeng25:
@Jitmisra you might want to think a bit deeper about this change.

1. after removing the calendar, should this page be renamed with a different title as there's no calendar? might as well rename the title to 'Events'.
2. after removing the calendar, should line12-19 be removed as there is no place to subscribe/add any events?

See #1850 for your reference, I'll mark it as a draft and you can continue your work in #1848.

website/docs/general/events.md
Jitmisra, line 1: ignore this commited update another

Jitmisra: Update events.md 1848 check this one

Jitmisra:
To simplify understanding, I've created a new PR with the same changes you provided at https://github.com/apache/apisix-website/pull/1851

yzeng25:
See #1851

PR #1949, opened 2025-08-15
fix: redirect old ingress next version links to new content

Baoyuantop:
Changes:

In #1948, I added a lot of redirect rules to avoid 404 ingress doc pages in search engine results. Later, I found that a type of rule was missing. This PR supplemented it.

Screenshots of the change:


SkyeYoung (dismissed):
May I ask if you have confirmed whether there are any omissions?

Baoyuantop:
> May I ask if you have confirmed whether there are any omissions?

The current changes are based on popular paths in Google search. It is not easy for us to add redirects for all 404 pages indexed by Google. If there is expired content later, we can submit it to Google.

.htaccess
```diff
@@ -77,6 +77,8 @@ Redirect 301 "/zh/blog/2021/11/30/use-apisix-ingress-in-kubesphere/httpbin.org/"
 Redirect 301 "/zh/blog/2021/08/25/Auth-with-Casbin-in-Apache-APISIX/" "/zh/blog/2021/08/18/auth-with-casbin-in-apache-apisix/"
 Redirect 301 "/zh/blog/2022/03/02/apisix-integration-graphql-plugin/" "/zh/blog/2022/03/02/apisix-integration-graphql/"
 
+RedirectMatch 301 "^/docs/ingress-controller/next/(.*)$" "/docs/ingress-controller/1.8.0/$1"
```
juzhiyuan: `next` directory should point to `2.x` I think?

Baoyuantop: The documentation in our previous "next" directory is equivalent to the current documentation in the "1.8.0" directory, with their directories and contents completely matching. If we switch to version 2.x, the document paths for the old and new versions will be significantly different, requiring additional logic to handle these complex relationships, or simply redirecting to the 2.x homepage. Given that each old version document has a tooltip and link to the new version at the top, I think this is acceptable.

kayx23 (dismissed):

.htaccess
Baoyuantop: Hi @juzhiyuan, please check this when you have time.

juzhiyuan: In Docusaurus version management, my understanding is that **Next** refers to the **latest unreleased content**. According to the current PR modification, when users access the **Next** directory, they will be redirected to **version 1.8.x**, which causes a conflict. I can accept this modification by hidding "Next" from the version options. (If @bzp2010 @kayx23 @moonming accept the current modification, then it's okay)

Baoyuantop: I agree with you that the current changes will affect the use of the existing next directory, and for long-term maintenance, preserving the existing next directory is necessary. I would modify the redirection rules to use more fine-grained rules, matching only specific paths rather than the entire next directory. This will increase the number of redirect records, but I think it is acceptable.

Baoyuantop: Hi @juzhiyuan, I have completed the modification and can check it again. At present, I have found some old broken links with high authority and try to find the matching content in the new version of the document. If there is no matching content, I will jump to the default main page.

@kayx23 If there is new document content, we can gradually remove these rules.

juzhiyuan (approved):

SkyeYoung (approved):

PR #2028, opened 2026-04-14
blog: add article on APISIX 3.16 dynamic rate limiting

moonming:
## Summary
- Add blog post: "What's New in Apache APISIX 3.16: Dynamic Rate Limiting for Your API Gateway"
- Covers two new features: multiple rules and variable support in `limit-count`, `limit-conn`, and `ai-rate-limiting` plugins
- Includes 6 complete, tested configuration examples (all validated against a live APISIX 3.16.0 instance)
- Covers use cases: multi-window rate limiting, per-tier quotas, multi-tenant isolation, concurrent connection limits, AI token budget management, and combined rules with variables

## Test plan
- [x] All 6 JSON configuration examples validated via APISIX 3.16.0 Admin API (201 Created)
- [x] Gateway requests verified rate limit headers are correctly returned
- [x] Article follows existing blog post format and frontmatter conventions

### DIFF
--- a/blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
+++ b/blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
@@ -110,10 +110,13 @@ Variable support lets you pull rate limiting parameters directly from the reques
 
 Suppose your authentication middleware injects an `X-Rate-Quota` header based on the user's subscription tier:
 
+> **Prerequisite**: this example uses `${consumer_name}` as the rate limit key, which requires an authentication plugin on the route so that the consumer identity is available at request time. Add `key-auth` (or any other auth plugin) to the route plugins and create a consumer before testing.
+
 ```json
 {
   "uri": "/api/v1/*",
   "plugins": {
+    "key-auth": {},
     "limit-count": {
       "rules": [
         {
@@ -178,10 +181,15 @@ Tenant A calling `/api/v1/users` and Tenant B calling the same endpoint get inde
 
 The `limit-conn` plugin also supports rules and variables, enabling dynamic concurrency control:
 
+> **Prerequisite**: the per-consumer rule uses `${consumer_name}`, which requires an authentication plugin. Add `key-auth` to the route plugins and create a consumer before testing.
+>
+> **Note**: plain constant strings (e.g. `"global"`) are not supported as `key` values in the `rules` array due to a current APISIX limitation — the rule is silently skipped at runtime. Use a variable expression that always resolves instead, such as `"${http_host ?? global}"`. A fix has been filed at [apache/apisix#13180](https://github.com/apache/apisix/issues/13180).
+
 ```json
 {
   "uri": "/api/v1/inference",
   "plugins": {
+    "key-auth": {},
     "limit-conn": {
       "default_conn_delay": 0.1,
       "rules": [
@@ -193,7 +201,7 @@ The `limit-conn` plugin also supports rules and variables, enabling dynamic conc
         {
           "conn": 100,
           "burst": 20,
-          "key": "global"
+          "key": "${http_host ?? global}"
         }
       ],
       "rejected_code": 503
@@ -211,6 +219,10 @@ The `limit-conn` plugin also supports rules and variables, enabling dynamic conc
 This limits each consumer to 5 concurrent connections while capping the total at 100 — preventing any single consumer from monopolizing backend capacity.
 
 ## AI Rate Limiting: Token Budget Management
+
+> **Prerequisite**: `ai-rate-limiting` must be used alongside the `ai-proxy` plugin. Without `ai-proxy`, the plugin is silently inactive — it relies on `ai-proxy` to populate `ctx.picked_ai_instance_name` and `ctx.ai_token_usage` at runtime. The configuration below shows `ai-rate-limiting` in isolation for clarity; in production, add your `ai-proxy` configuration to the same route.
+>
+> **Note**: the global cap rule uses `"${http_host ?? global}"` instead of a plain `"global"` string. See the note in Example 3 for the reason.
 
 For AI gateway use cases, the `ai-rate-limiting` plugin combines multiple rules with variable support for fine-grained token budget control:
 
@@ -236,7 +248,7 @@ For AI gateway use cases, the `ai-rate-limiting` plugin combines multiple rules 
         {
           "count": 1000000,
           "time_window": 60,
-          "key": "global",
+          "key": "${http_host ?? global}",
           "header_prefix": "global"
         }
       ],
@@ -264,10 +276,13 @@ As AI API costs scale directly with token usage, this kind of layered budget con
 
 The real power emerges when you combine both features. Here is a complete example for an API platform with tiered pricing:
 
+> **Prerequisite**: add an authentication plugin (e.g. `key-auth`) to the route so that `${consumer_name}` is populated at runtime. The global cap rule uses `"${http_host ?? global}"` instead of a plain `"global"` string — see the note in Example 3 for the reason.
+
 ```json
 {
   "uri": "/api/v1/*",
   "plugins": {
+    "key-auth": {},
     "limit-count": {
       "rules": [
         {
@@ -285,7 +300,7 @@ The real power emerges when you combine both features. Here is a complete exampl
         {
           "count": 100000,
           "time_window": 60,
-          "key": "global",
+          "key": "${http_host ?? global}",
           "header_prefix": "global"
         }
       ],
--- a/blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
+++ b/blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
@@ -14,7 +14,7 @@ keywords:
   - Multi-Tenant
   - Token Budget
 description: Apache APISIX 3.16 introduces dynamic rate limiting with multiple rules and variable support across limit-count, limit-conn, and ai-rate-limiting plugins, enabling context-aware, per-tier, and multi-tenant rate limiting in a single route configuration.
-tags: [Products]
+tags: [Community]
 ---
 
 Rate limiting is one of the most critical capabilities in any API gateway. Yet for years, most gateways — including APISIX — have treated it as a static, one-size-fits-all configuration: set a number, set a time window, done.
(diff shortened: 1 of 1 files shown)


blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,328 @@
...
+```bash
+# Docker
+docker pull apache/apisix:3.16.0
+
+# Helm
+helm repo update
+helm upgrade apisix apisix/apisix --set image.tag=3.16.0
+```
+
+Check the full documentation:
```
Baoyuantop, line 328:
```suggestion
Check the full documentation:

```

Baoyuantop (dismissed):

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+        {
+          "count": "${http_x_rate_quota ?? 100}",
+          "time_window": 60,
+          "key": "${consumer_name}"
+        }
+      ],
+      "rejected_code": 429
+    }
+  },
+  "upstream": {
```
Yilialinn, line 129: **Bug: this example returns `500 Internal Server Error` at request time when no consumer is authenticated.**

Tested on APISIX 3.16.0. Error log:
```
[lua] init.lua:456: phase_func(): failed to get rate limit rules: nil
limit-count exits with http status code 500
```

Root cause: in `limit-count/init.lua` `get_rules()`, after resolving the key string, there is a check:
```lua
local key, _, n_resolved = core.utils.resolve_var(rule.key, ctx.var)
if n_resolved == 0 then
    goto CONTINUE  -- rule is silently skipped
end
```
`n_resolved` counts how many `${}` variables were successfully resolved to a non-nil value. When `consumer_name` is nil (no authenticated consumer), `n_resolved` stays `0` and the rule is skipped. When all rules are skipped, the rules list is empty and APISIX returns 500.

**This is a code-level bug in APISIX** — `n_resolved == 0` was likely intended to skip rules where the key variable is missing, but it also incorrectly skips rules whose key variables resolve to nil. A separate issue should be filed against `apache/apisix`.

**To make this example work as written**, the route must include an authentication plugin so that `consumer_name` is populated:
```json
"plugins": {
  "key-auth": {},
  "limit-count": { ... }
}
```
And a consumer with that auth plugin must exist. The blog should either add this prerequisite or use a key that is always available (e.g., `"${remote_addr}"`) for the basic illustration.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+        {
+          "conn": 5,
+          "burst": 2,
+          "key": "${consumer_name}"
+        },
+        {
+          "conn": 100,
+          "burst": 20,
+          "key": "global"
+        }
```
Yilialinn, line 199: **Bug: same `n_resolved == 0` issue — `${consumer_name}` key causes this rule to be skipped when no consumer is authenticated, resulting in `500`.**

Tested on APISIX 3.16.0. Error log:
```
[lua] init.lua:208: phase_func(): failed to get limit conn rules: nil
limit-conn exits with http status code 500
```

Same root cause as the `limit-count` case: `limit-conn/init.lua` has the identical `n_resolved == 0` guard. The `${consumer_name}` rule is skipped when there is no consumer context.

**Fix needed**: add an auth plugin (e.g., `key-auth`) to the route so `consumer_name` is populated at request time.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+      ],
+      "rejected_code": 503
+    }
+  },
+  "upstream": {
```
Yilialinn, line 204: **Bug: a plain string key like `"global"` also triggers the `n_resolved == 0` skip and is never applied.**

This is the second part of the `limit-conn` bug. Since `"global"` contains no `${}` variable expression, `core.utils.resolve_var("global", ctx.var)` returns `n_resolved = 0`. The guard `if n_resolved == 0 then goto CONTINUE` then skips this rule too.

In this example both rules are skipped → empty rules list → `500`.

**This is also a code-level bug in APISIX**: constant string keys should be valid and usable directly. The `n_resolved == 0` check should only skip a rule when the key *contains* a variable expression that failed to resolve, not when the key is a plain constant.

**Workaround until the APISIX bug is fixed**: replace the bare constant with a variable expression that always resolves:
```json
"key": "${http_host ?? global}"
```
`http_host` is present on every request, so `n_resolved = 1`. The resulting key value is the Host header (effectively a per-service shared counter), or `"global"` as fallback if the header is absent.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+{
+  "uri": "/v1/chat/completions",
+  "plugins": {
+    "ai-rate-limiting": {
+      "limit_strategy": "total_tokens",
+      "rules": [
+        {
+          "count": 10000,
+          "time_window": 60,
+          "key": "${consumer_name}_per_minute",
```
Yilialinn, line 229: **Bug: `ai-rate-limiting` is silently non-functional when used without `ai-proxy`.**

Tested on APISIX 3.16.0. The route accepts requests and returns `502` (upstream unreachable), but the rate limiting plugin does absolutely nothing — no rate limit headers, no token counting, no rejection.

Root cause: `ai-rate-limiting`'s `access()` begins with:
```lua
local ai_instance_name = ctx.picked_ai_instance_name
if not ai_instance_name then
    return  -- silently exits, no limiting applied
end
```
`ctx.picked_ai_instance_name` is only set by the `ai-proxy` (or `ai-proxy-multi`) plugin. Without it, the entire plugin is a no-op.

**The example as written is incomplete.** It must include an `ai-proxy` plugin configuration on the same route for `ai-rate-limiting` to have any effect. Suggest adding a note:

> **Prerequisite**: `ai-rate-limiting` must be used alongside the `ai-proxy` plugin. It relies on `ai-proxy` to populate the AI instance context (`ctx.picked_ai_instance_name`) and token usage (`ctx.ai_token_usage`). Without `ai-proxy`, the plugin is silently inactive.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+        {
+          "count": 1000000,
+          "time_window": 60,
+          "key": "global",
+          "header_prefix": "global"
+        }
+      ],
+      "rejected_code": 429
+    }
+  },
```
Yilialinn, line 247: **Bug: `"key": "global"` is silently skipped here too (same `n_resolved == 0` issue).**

`ai-rate-limiting` delegates to `limit-count` internally for the `rules` path. The same `n_resolved == 0` guard applies, so the `"global"` constant key rule is never enforced.

Combined with the missing `ai-proxy` issue on the same example, the global capacity cap described in point 3 of the explanation below this block never takes effect.

**Workaround**: `"key": "${http_host ?? global}"` (same fix as the `limit-conn` example).

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+        {
+          "count": "${http_x_burst_quota ?? 10}",
+          "time_window": 1,
+          "key": "${consumer_name}_per_second",
+          "header_prefix": "burst"
+        },
+        {
+          "count": "${http_x_sustained_quota ?? 500}",
+          "time_window": 60,
+          "key": "${consumer_name}_per_minute",
```
Yilialinn, line 285: **Bug: `${consumer_name}_per_second` key fails when no consumer is authenticated — same `n_resolved == 0` issue.**

Without an auth plugin on the route, `consumer_name` is nil. The rule is skipped, and since rule 3 (`"global"`) is also skipped (see next comment), all rules are dropped → `500`.

**Fix**: add `"key-auth": {}` (or another auth plugin) to the route plugins.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
```diff
@@ -0,0 +1,329 @@
...
+        {
+          "count": 100000,
+          "time_window": 60,
+          "key": "global",
+          "header_prefix": "global"
+        }
+      ],
+      "rejected_code": 429
+    }
+  },
```
Yilialinn, line 297: **Bug: `"key": "global"` is silently skipped — same `n_resolved == 0` issue as the `limit-conn` example.**

Even if rules 1 and 2 were working (with a consumer), this global safety cap rule would never be applied because the constant string `"global"` produces `n_resolved = 0`.

This means the "static global safety cap" described in the paragraph below this block is never actually enforced, which defeats a key part of the example's purpose.

**Workaround**: `"key": "${http_host ?? global}"` until the upstream APISIX bug is fixed.

Yilialinn (commented):
Reviewed all 6 configuration examples by running them against a live APISIX 3.16.0 instance. All examples pass Admin API schema validation (201 Created). However, 3 examples fail at runtime and 1 is silently non-functional. Details are inline below.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
Yilialinn: Please check if these two pieces of advice on `limit-count` and `limit-conn` plugins are reasonable to apply @nic-6443 :
image

nic-6443: This is the expected behavior, not a bug. You used the Consumer Name in the configured Rules but did not set up any authentication plugins, so naturally no Rule can be matched. When no Rule is matched, a 500 error will occur.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
nic-6443: The rules.key only supports the use of variables and does not support constants; this is by design.

blog/en/blog/2026/04/14/apisix-3.16-dynamic-rate-limiting.md
nic-6443: ditto

Yilialinn (approved):

Baoyuantop (approved):

PR #888, opened 2022-02-18
fix: update contributing guide

pottekkat:
Signed-off-by: Navendu Pottekkat <navendupottekkat@gmail.com>

Partial completion of #887

View changes: [/docs/general/contributor-guide/](https://deploy-preview-888--apache-apisix.netlify.app/docs/general/contributor-guide/)
 
Changes:

1. Fixed typos
2. Added hyperlinks
3. Improved readability by restructuring sentences

PR #2063, opened 2026-06-26
fix(seo): a11y + CLS polish (homepage contrast/headings + Showcase logos)

moonming:
## Summary

Homepage accessibility polish — the P2 items from the homepage-rewrite audit (now that `fix/homepage-layout` has landed):

- **WCAG AA contrast** — darken low-contrast text:
  - secondary greys `#777` / `#8a8a8a` / `#888` → `#6b6b6b` (~5.3:1)
  - CTA red text `#e8433e` → `#c7352f` (the **existing hover shade** in the palette, ~5.3:1) on the Pathways / Integrations / AI-gateway CTAs and the comparison APISIX column
  - left the large bold hero stat value (passes AA-large at 3:1) and the decorative `aria-hidden` flow arrow unchanged
- **Heading hierarchy** — Pathways card titles `h3` → `h2`, fixing the homepage `h1`→`h3` skip. Styling is class-driven (`.pathways__title`), so there is **no visual change**.

8 lines changed; no layout or structural changes.

> The CTA change keeps the brand red, just an AA-compliant darker shade for normal-size text (`#c7352f` is already the hover color). If you'd rather keep the brighter `#e8433e` on CTAs, that half is trivial to revert.

## Out of scope
- The 2 dimensionless `` the CWV check flagged are on `/showcase/` and `/plugins/`, not the homepage (the homepage Integrations logos already have explicit `width`/`height`).
- Real CWV numbers still need a PSI API key or a Lighthouse run.

## Test plan
- [x] Diff verified: exactly the 8 intended color/heading changes, no collateral
- [x] New colors are objectively WCAG AA (`#6b6b6b` 5.3:1, `#c7352f` ~5.3:1); class-driven `h2` = no visual regression
- [ ] CI deploy build green

### DIFF
--- a/website/src/components/sections/Pathways.tsx
+++ b/website/src/components/sections/Pathways.tsx
@@ -53,7 +53,7 @@ const Pathways: FC = () => (
   <section className="pathways" aria-label="Get started with Apache APISIX">
     {PATHWAYS.map((pathway) => (
       <Link className="pathways__card" to={pathway.to} key={pathway.id}>
-        <h3 className="pathways__title">{pathway.title}</h3>
+        <h2 className="pathways__title">{pathway.title}</h2>
         <p className="pathways__desc">{pathway.description}</p>
         <span className="pathways__cta">
           {pathway.cta}
--- a/website/src/css/landing-sections/ai-gateway-home.scss
+++ b/website/src/css/landing-sections/ai-gateway-home.scss
@@ -84,7 +84,7 @@
 }
 
 .ai-home__provider--more {
-  color: #888;
+  color: #6b6b6b;
 }
 
 .ai-home__grid {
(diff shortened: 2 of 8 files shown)


moonming:
Added a second commit: reserve a fixed 100×40 `object-fit: contain` box for the **Showcase** user-logo wall (+ matching `width`/`height` attrs) so the variable-aspect logos no longer cause layout shift, without distortion.

**Plugin Hub logos left intentionally:** they use a `250×180` max and are dynamically sourced (variable aspect), so a correct CLS fix means box-reserving 250×180 — which would balloon small logos with whitespace and visibly change the hub just reworked in #2061. Not worth it for below-the-fold, lazy-loaded images; flagging for a deliberate design pass if desired.

PR #2078, opened 2026-07-26
feat(docs): serve the latest-version docs from the static build (wave 3)

moonming:
## What

Serves the **latest-version** project docs from the static Astro build. **Version archives stay on Docusaurus** — this PR does not touch them.

Scope is 486 pages (243 per locale) across the seven documented projects, not the full 3410-page docs tree.

## Why the split works

Production already separates the two: the current release lives at version-less URLs (`/docs/apisix/plugins/cors/`) while archives live under `3.16/`, `next/`, etc. Migrating only the version-less layer avoids the versioning problem entirely.

## How

**Content**: CI sparse-clones `docs/` from the seven upstream Apache repos at the **same refs `scripts/sync-docs.js` uses** — the newest release series from `config/apisix-versions.js` for apisix, and for each sub-project the newest release ref detected on the remote (`v<ver>` for ingress-controller, `release/<ver>` elsewhere), ordered with semver. Not the default branch: master is what production publishes at `/docs/<project>/next/`, and shipping it at the version-less URLs would document unreleased behaviour as the current release.

**Overlay — per-page, not per-subtree.** Under `docs/<project>/` the version directories sit *beside* the latest-version pages, so a subtree swap would delete the archive. Only paths Astro produced are copied, and the step asserts `3.16/` and `next/` are still Docusaurus.

**Archive links**: the archives remain a Docusaurus SPA whose version dropdown and "older version" banner point at the version-less URLs this PR replaces. Without a real navigation React would client-route to its own stale render, so those links get `target="_self"` after the overlay — the same trick the repo already uses for `pathname://` links.

**Parity gate**: fails the deploy if a version-less page Docusaurus built is missing from the Astro build. Two documented exclusions: version directories (by design) and `docs/*/tags/` — Docusaurus emits those empty (one self-link, absent from the sitemap, linked from no doc page); the per-page swap leaves them in place so the URLs keep resolving.

**Search**: Algolia DocSearch against the index the site already uses (`apache_apisix`), on docs pages only. The homepage, blog, and learning center stay zero-JS.

## Verification

Simulated against the live `asf-site` tree:

| | |
|---|---|
| Pages | 3410 → **3410** |
| Deleted / added | **0 / 0** |
| Archived version pages compared | **2908, all byte-identical** |
| Parity gate | passes for both locales |

Relative `.md` links now resolve against the linking document's own directory (previously they collapsed to the project root — 403 broken links across 141 pages). 29 broken links remain, all pre-existing upstream authoring errors that production renders identically.

`canonical` frontmatter pointing at docs.api7.ai is preserved on all 160 pages that carry it, `og:url` stays on apisix.apache.org (matching production), and both are re-asserted after the overlay rather than only against the Docusaurus build the overlay replaces.

## Known gaps

- The kapa.ai assistant widget that Docusaurus injects on docs pages is not carried over; it remains on the archives, so the two differ until it is added back.
- DocSearch runs without contextual filtering (`docsearch:version`/`language` metas), so results can include archived versions and the other locale.
- The search box is hidden below 996px.

## Follow-ups

Version archives remain Docusaurus, so that build stays in the pipeline. Retiring it is a later decision.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -266,24 +266,26 @@ jobs:
               echo "NON-HTML files not in the Astro tree and not in the feed-carry list:"; echo "$unhandled"; fail=1
             fi
           done
-          # Latest-version docs: every version-less page Docusaurus produced
-          # must exist in the Astro build, or the swap would leave a stale
-          # Docusaurus page behind a URL we claim to own. Versioned dirs are
-          # excluded — they stay Docusaurus by design. So are docs/*/tags/
-          # index pages: Docusaurus emits them empty (one self-link, absent
-          # from the sitemap, linked from no doc page). The per-page swap
-          # leaves them in place, so the URLs keep resolving.
+          # Latest-version docs: Docusaurus no longer builds them (see
+          # onlyIncludeVersions in doc/docusaurus.config.js), so the reference
+          # is the LIVE SITE. Every version-less docs URL currently published
+          # must exist in the Astro build, or this deploy would 404 a page that
+          # works today. Versioned dirs are excluded — they stay Docusaurus by
+          # design — as are docs/*/tags/ index pages, which Docusaurus emits
+          # empty (one self-link, absent from the sitemap, linked from no doc
+          # page) and which the per-page swap leaves in place.
+          git fetch --depth 1 origin asf-site
           for docs_root in docs zh/docs; do
-            [ -d "website/build/$docs_root" ] || continue
-            missing=$(cd "website/build/$docs_root" && find . -name index.html \
-              | sed 's|^\./||' \
+            missing=$(git ls-tree -r FETCH_HEAD --name-only -- "$docs_root" \
+              | grep '/index\.html$' \
+              | sed "s|^$docs_root/||" \
               | grep -vE '(^|/)([0-9]+\.[0-9]+|next|v[0-9][^/]*)/' \
               | grep -vE '(^|/)tags/index\.html$' \
               | sort | while read -r f; do
                 [ -f "$ROOT/next/dist/$docs_root/$f" ] || echo "$docs_root/$f"
               done)
             if [ -n "$missing" ]; then
-              echo "MISSING latest-version docs in the Astro build:"; echo "$missing"; fail=1
+              echo "Published docs URLs missing from the Astro build:"; echo "$missing"; fail=1
             fi
           done
           exit $fail
@@ -344,34 +346,31 @@ jobs:
           # …and the latest-version pages must now be the Astro build.
           grep -q '/_astro/' website/build/docs/apisix/plugins/cors/index.html
           grep -q '/_astro/' website/build/zh/docs/apisix/plugins/cors/index.html
-          # The archives are still a Docusaurus SPA, and their version dropdown
-          # and "older version" banner link to the version-less URLs we just
-          # replaced. Without a real navigation, React would client-route to
-          # its own stale render of those URLs instead of fetching the Astro
-          # page. target="_self" forces a document load (same trick the repo
-          # already uses for pathname:// links in the locale dropdown).
-          node -e '
+          # The archives remain a Docusaurus SPA whose version dropdown and
+          # "older version" banner point at the version-less URLs. Docusaurus
+          # is configured (doc/docusaurus.config.js, onlyIncludeVersions) to
+          # stop building the newest version, so those URLs are no longer in
+          # its route manifest and a click leaves the SPA for the Astro page.
+          # Assert that manifest property directly — patching the serialized
+          # HTML would not help, because React Router intercepts clicks after
+          # hydration regardless of the anchor's target.
+          if ! node -e '
             const fs = require("fs"), path = require("path");
-            const roots = ["website/build/docs", "website/build/zh/docs"];
-            // /docs/<project>/<page…>/ with no version segment right after the
-            // project — i.e. exactly the URLs the Astro build now owns.
-            const re = /(<a\b[^>]*?\bhref="\/(?:zh\/)?docs\/[a-z0-9-]+\/(?!(?:[0-9]+\.[0-9]+|next|v[0-9])\/)[^"]*")/g;
-            let patched = 0;
-            const walk = (d) => fs.existsSync(d) && fs.readdirSync(d, { withFileTypes: true }).forEach((e) => {
-              const p = path.join(d, e.name);
-              if (e.isDirectory()) return walk(p);
-              if (e.name !== "index.html") return;
-              // Only the archived pages need this; the Astro pages are not a SPA.
-              const html = fs.readFileSync(p, "utf8");
-              if (!html.includes("docusaurus")) return;
-              const out = html.replace(re, (m) => (m.includes("target=") ? m : `${m} target="_self"`));
-              if (out !== html) { fs.writeFileSync(p, out); patched += 1; }
-            });
-            roots.forEach(walk);
-            console.log(`archive pages patched for real navigation: ${patched}`);
-          '
-          # An archive page must now hard-navigate to the migrated URL.
-          grep -q 'href="/docs/apisix/plugins/cors/" target="_self"' website/build/docs/apisix/3.16/plugins/cors/index.html
+            const dir = "website/build/assets/js";
+            const files = fs.existsSync(dir) ? fs.readdirSync(dir).filter((f) => /^main\..*\.js$/.test(f)) : [];
+            if (!files.length) { console.error("no main bundle found"); process.exit(1); }
+            const bundle = fs.readFileSync(path.join(dir, files[0]), "utf8");
+            const claimed = ["/docs/apisix/plugins/cors/", "/docs/apisix/getting-started/"]
+              .filter((u) => bundle.includes(`"${u}"`));
+            if (claimed.length) {
+              console.error("Docusaurus still owns migrated routes:", claimed.join(", "));
+              process.exit(1);
+            }
+            console.log("route manifest: no version-less docs routes — archive links hard-navigate");
+          '; then
+            echo "The Docusaurus SPA still routes the migrated docs URLs client-side."
+            exit 1
+          fi
           # The cross-site canonical contract asserted earlier in this job runs
           # against the Docusaurus build; re-assert it on what actually ships.
           for f in website/build/docs/apisix/plugins/cors/index.html \
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -165,8 +165,21 @@ jobs:
         working-directory: next
         run: |
           npm ci
+          # Project docs live in the upstream Apache repos. Sparse, depth-1
+          # clones of just docs/ keep this to a few seconds each; the sync
+          # script reads .sync/<repo>/docs/{en,zh}/latest and skips any repo
+          # that is absent, so a fetch failure degrades to "no pages for that
+          # project" — which the parity gate below then catches.
+          mkdir -p .sync
+          for repo in apisix apisix-ingress-controller apisix-helm-chart \
+                      apisix-docker apisix-java-plugin-runner \
+                      apisix-go-plugin-runner apisix-python-plugin-runner; do
+            git clone --depth 1 --filter=blob:none --sparse -q \
+              "https://github.com/apache/$repo.git" ".sync/$repo" || continue
+            (cd ".sync/$repo" && git sparse-checkout set docs) || true
+          done
           # Blog/learning-center/articles markdown is synced from this very
-          # checkout (WEBSITE_REPO); project docs stay un-synced until wave 3.
+          # checkout (WEBSITE_REPO); project docs come from .sync/ above.
           WEBSITE_REPO="$GITHUB_WORKSPACE" node scripts/sync-content.mjs
           npx astro build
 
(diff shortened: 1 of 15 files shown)


juzhiyuan (dismissed):

juzhiyuan (dismissed):

.github/workflows/deploy.yml
```diff
@@ -229,6 +315,72 @@ jobs:
...
+            const re = /(<a\b[^>]*?\bhref="\/(?:zh\/)?docs\/[a-z0-9-]+\/(?!(?:[0-9]+\.[0-9]+|next|v[0-9])\/)[^"]*")/g;
+            let patched = 0;
+            const walk = (d) => fs.existsSync(d) && fs.readdirSync(d, { withFileTypes: true }).forEach((e) => {
+              const p = path.join(d, e.name);
+              if (e.isDirectory()) return walk(p);
+              if (e.name !== "index.html") return;
+              // Only the archived pages need this; the Astro pages are not a SPA.
+              const html = fs.readFileSync(p, "utf8");
+              if (!html.includes("docusaurus")) return;
+              const out = html.replace(re, (m) => (m.includes("target=") ? m : `${m} target="_self"`));
```
LiteSun: Adding `target="_self"` does not force a document navigation in the hydrated Docusaurus app. Docusaurus renders internal URLs through React Router, and React Router still intercepts clicks when the target is absent or `_self`, calls `preventDefault()`, and performs client-side history navigation. Patching the generated HTML also does not change the link props React hydrates with. The archive dropdown/banner can therefore still render Docusaurus's stale versionless page instead of fetching the Astro page. Please use a URL/link form that Docusaurus treats as external, or add an explicit hard-navigation handler, and verify the behavior after hydration rather than only asserting the serialized HTML.

next/scripts/generate-md-twins.mjs
```diff
@@ -0,0 +1,160 @@
...
+ * back to the English text where no translation exists), so a source file can
+ * legitimately map to two URLs.
+ */
+const COLLECTIONS = [
+  ['blog-en', ['/blog']],
+  ['blog-zh', ['/zh/blog']],
+  ['learning-center', ['/learning-center', '/zh/learning-center']],
+  ['articles', ['/articles', '/zh/articles']],
+  ['docs-general', ['/docs/general', '/zh/docs/general']],
+  ['docs-apisix-en', ['/docs/apisix']],
```
LiteSun: This mapping does not cover the English fallback used by the Chinese APISIX docs. When a page has no Chinese translation, the build still publishes an English-backed page under `/zh/docs/apisix/...`, but `docs-apisix-en` is only checked against `/docs/apisix`. In the release build this leaves 21 published Chinese-locale HTML pages without the promised `index.md` twin. Please include the Chinese prefix as a fallback for this collection, or add a parity assertion that every migrated HTML page has a twin.

next/scripts/generate-md-twins.mjs
```diff
@@ -0,0 +1,160 @@
...
+  '',
+  'Every page below is available as Markdown — append `index.md` to any page URL.',
+  '',
+  ...section('Documentation', group(en, '/docs/')),
+  ...section('Learning center', group(en, '/learning-center/')),
+  ...section('Blog', group(en, '/blog/')),
+  ...section('Articles', group(en, '/articles/')),
+  ...section('中文文档', group(zh, '/docs/')),
+  ...section('中文学习中心', group(zh, '/learning-center/')),
+  ...section('中文博客', group(zh, '/blog/')),
```
LiteSun, line 166: The `articles` collection generates both English and Chinese-locale twins, but the index has no Chinese articles section. The generated `/zh/articles/.../index.md` files (13 in the current build) are therefore omitted from `/llms.txt`. Please add the corresponding section so the index covers the twins this script publishes.

LiteSun (commented):
I re-checked these findings against the staged Wave 3 boundary. Two rollout regressions are not on newly added lines, but they become production behavior when this PR overlays the current-version docs with the Astro output:

- The sidebar renderer stops after three levels and filters the last level to nodes with an `id`. A nested category such as APISIX 3.17's `Plugins > Observability > Tracers/Metrics/Loggers` therefore renders empty and drops the links to the 25 documents below it. The relevant renderer is [here](https://github.com/apache/apisix-website/blob/38dab2073ece853894ae7e0072b36ac8d53d3b26/next/src/layouts/DocPage.astro#L54-L65). Please render the sidebar recursively, or otherwise preserve all configured descendants before switching these pages.
- The project-doc routes still build Edit links with `master` and `entry.id` ([source](https://github.com/apache/apisix-website/blob/38dab2073ece853894ae7e0072b36ac8d53d3b26/next/src/pages/docs/%5Bproject%5D/%5B...id%5D.astro#L30)). This PR now syncs release refs, so the link can target different content from the page being viewed; slugs can also differ from the source path. For example, the ingress-controller annotation page points to `master/docs/en/latest/reference/apisix-ingress-controller/annotation.md`, while its source is `v2.1.0/docs/en/latest/reference/annotations.md`. Please carry the resolved ref and source-relative path into the edit URL.

The migrated latest pages also show a version label without links back to `next` or the retained archives. If restoring the selector is intentionally deferred to a later wave, please list that explicitly under the known gaps; otherwise this rollout makes the archives undiscoverable from the latest-doc UI.

The Markdown-twin commit is separable from the core Wave 3 switch. If it remains in this PR, the inline completeness findings below should be addressed because that commit promises a twin for every content page and a complete index.

moonming:
Thanks — all six points were real, and the first one showed my fix was based on a wrong assumption. Addressed in 442343c71.

**Client-side routing.** You're right that `target="_self"` does nothing here: React Router intercepts the click after hydration regardless of target, and patching the serialized HTML never reaches the props React hydrates with. I've dropped that approach. Docusaurus now stops building the newest version (`onlyIncludeVersions` in `doc/docusaurus.config.js`), so the version-less URLs leave its route manifest entirely and a click from an archive page is a real navigation. The deploy asserts that against the built `main.*.js` bundle instead of against markup.

That change had a consequence worth flagging: with Docusaurus no longer emitting those pages, the docs parity gate had nothing left to compare and would have passed vacuously. It now checks the Astro build against the URLs currently published on `asf-site`, so a page that works today cannot 404 after a deploy.

**Sidebar depth.** Confirmed and fixed — 25 links under `Plugins > Observability > {Tracers,Metrics,Loggers}` were being dropped. Replaced the fixed-depth markup with a recursive component; `zipkin` and the other 24 are back.

**Edit links.** Fixed. `sync-content.mjs` now records the ref each project was cloned from (`release/3.17`, `v2.1.0`, `release/0.6.0`, …) and the pages build the link from that ref plus `pathId` — the source-relative path — so a frontmatter slug no longer points at a file that doesn't exist. Your ingress-controller example now resolves to a real source file.

**Markdown twins — zh fallback.** Fixed: the English APISIX collection is now also offered the `/zh/` prefix, covering the 192 zh pages that fall back to English. Rather than rely on that mapping staying correct, I added a parity check that fails the build if any content page lacks a twin.

**Markdown twins — zh articles in the index.** Fixed; the section is there and the 13 pages are indexed.

**Version picker.** Not deferred — added. The migrated pages now carry a `details/summary` picker listing the seven archived versions plus `next`, in both locales, with no JavaScript.

On separability: happy to split the Markdown-twin commit into its own PR if you'd prefer to review the Wave 3 switch on its own.

.github/workflows/deploy.yml
LiteSun: Thanks for replacing the `target="_self"` patch, but the new `onlyIncludeVersions` approach does not resolve this yet. The generated versions JSON contains release names only; Docusaurus injects `current` separately, chooses the first included release as `lastVersion`, and assigns that version the empty (version-less) path. For APISIX, `all.slice(1)` therefore excludes both `current` and 3.17, makes 3.16 own the version-less routes, and stops producing both `/3.16/` and `/next/`. The archive-preservation assertions later in this workflow will fail, while the route manifest still owns the migrated URLs with 3.16 content. For every sub-project, `SUBPROJECT_VERSIONS_TO_KEEP` is 1, so the helper returns `undefined` and their version-less routes remain unchanged. Please configure explicit paths for the retained versions/current route and verify all seven plugin manifests plus the exact `3.16/` and `next/` outputs.

next/src/lib/content.ts
```diff
@@ -95,6 +95,22 @@ const docsPythonEn = import.meta.glob('/content/docs-python-plugin-runner-en/**/
...
+
+/**
+ * "Edit this page" URL for an upstream project doc. Uses the ref the content
+ * was actually synced from — sub-projects are cloned at a release tag, and
+ * default branches differ (master vs main) — and `pathId`, the source-relative
+ * path, since a frontmatter slug can move the URL away from the filename.
+ */
+export function docEditUrl(project: string, repo: string, pathId: string, locale: Locale): string {
+  const ref = docRefs[project] ?? 'master';
+  return `https://github.com/apache/${repo}/edit/${ref}/docs/${locale}/latest/${pathId}.md`;
```
LiteSun: Using the rendered locale as the source locale still produces broken Edit links for fallback pages. `getApisixDocs('zh')` and `getSubprojectDocs(..., 'zh')` deliberately use the English module when a translation is absent, but these routes still call this helper with `zh`. In the release build, 57 Chinese-locale pages have no Chinese source: for example `/zh/docs/apisix/aws/` links to `release/3.17/docs/zh/latest/aws.md` and `/zh/docs/helm-chart/apisix/` links to `master/docs/zh/latest/apisix.md`; both paths are missing while their English files exist. Please retain the effective source locale/path on `DocEntry` and build the Edit URL from those values.

next/src/pages/docs/[project]/[...id].astro
```diff
@@ -27,5 +27,5 @@ const repo = SUBPROJECTS[project].repo;
   sidebar={sidebar}
   titleById={titleById}
   urlById={urlById}
-  editUrl={`https://github.com/apache/${repo}/edit/master/docs/en/latest/${entry.id}.md`}
+  editUrl={docEditUrl(project, repo, entry.pathId, 'en')}
```
Comment (LiteSun) [inline next/src/pages/docs/[project]/[...id].astro]: The version picker was added only to the APISIX-specific routes. This generic route (and its Chinese counterpart) still passes no `versionLabel`, `archivedVersions`, or `versionBase`, so all six sub-projects lose their route back to the retained `next/` tree. The release build has 72 affected pages across the two locales. Please supply the per-project version metadata here as well, or explicitly document this as a deferred gap.

next/scripts/generate-md-twins.mjs
```diff
@@ -0,0 +1,199 @@
...
+  ['blog-zh', ['/zh/blog']],
+  ['learning-center', ['/learning-center', '/zh/learning-center']],
+  ['articles', ['/articles', '/zh/articles']],
+  ['docs-general', ['/docs/general', '/zh/docs/general']],
+  // The zh APISIX docs fall back to the English source where no translation
+  // exists, so the English collection must also be offered the zh prefix —
+  // resolveUrl only writes where a page was actually built, and the zh
+  // collection is processed after, overwriting the fallback where a real
+  // translation exists.
+  ['docs-apisix-en', ['/docs/apisix', '/zh/docs/apisix']],
```
LiteSun, line 41: Offering the English collection at the Chinese prefix writes and records an English fallback for every built Chinese APISIX page, not only pages without a translation. The Chinese collection then overwrites the file but pushes the same URL into `written` again, so `/llms.txt` contains both English and Chinese metadata for translated pages. The release build produced 1,186 unique twin files but 1,359 index entries, with 173 duplicate URLs. Please key the collected metadata by URL and let the translated entry replace the fallback; the parity check should also validate the actual links emitted to `llms.txt` rather than a Set derived from `written`.

LiteSun (commented):
The original ref/path edit-link issue is fixed for pages whose source locale matches the rendered locale, and the two Markdown completeness threads have been resolved. The inline comments below cover the remaining fallback, picker, and index-output gaps.

There is also one rollout regression in unchanged sidebar-normalization code: `getApisixSidebar()` and `getSubprojectSidebar()` only preserve strings, docs, and categories. A `type: "link"` item becomes a category with no children and is filtered out by `SidebarNodes`, removing APISIX's `CODE_STYLE`/`CHANGELOG` links and the ingress-controller `CHANGELOG` link. Both functions also always select the English `config.json`, so Chinese pages render English category labels even when a Chinese sidebar config exists. Please preserve link nodes and select the config by locale before enabling these pages.

moonming:
All four correct again — thanks for digging into the Docusaurus internals. Addressed in f1b2399c4.

**Version routing.** You're right, and my fix was worse than the problem: `versions.json` holds release names only (`current` is injected separately by `includeCurrentVersion`), so `all.slice(1)` dropped 3.17, promoted 3.16 into the version-less path with older content, and stopped building both `3.16/` and `next/`. I confirmed this against the beta.6 sources — `getDefaultLastVersionName` picks the first non-`current` entry, and `isLast` is what gets the empty path.

Every version now keeps building; the newest release just gets an explicit `path` so it stops owning the version-less URLs. Sub-projects are covered too — as you noted, the old helper returned `undefined` for all six, so their routes were never touched.

One consequence worth calling out explicitly: this publishes a **new** URL set — `/docs/apisix/3.17/` and the equivalent per sub-project — that the live site does not have today. Nothing is removed; the newest release simply becomes addressable by version like every other one. If you'd rather it not be reachable at all, `banner`/`noIndex` on that version entry is the lever, but I'd argue an addressable current release is the more useful default.

**Fallback edit links.** Fixed properly this time. `DocEntry` now carries `sourceLocale` — the locale of the file the page was actually rendered from — and the edit URL is built from that rather than the rendered locale. Your two examples now resolve: `/zh/docs/apisix/aws/` → `release/3.17/docs/en/latest/aws.md`, `/zh/docs/helm-chart/apisix/` → `master/docs/en/latest/apisix.md`, while a genuinely translated page like `/zh/docs/apisix/plugins/cors/` still points at `docs/zh/latest/`.

**Sub-project version picker.** Added to the generic routes; all six now link back to `next/`.

**Twin index duplication.** Fixed — the index is keyed by URL, so the translated entry replaces the fallback instead of appending, and the parity check now reads the links actually emitted into `llms.txt` rather than a Set derived from the write log. I also added an assertion that fails the build on any duplicate entry, which immediately earned its keep: it caught a second bug where `/docs/general/blog/` was filed under both the docs and blog sections because grouping used a substring test. Files and index entries are now 1:1 at 1,186.

juzhiyuan (approved):

LiteSun (approved):

PR #2084, opened 2026-07-28
fix(ci): check a generated blog Markdown twin

moonming:
Related: #2083

Changes:

- Replace the impossible blog section index.md assertion with a check for a generated blog post Markdown twin.
- Keep the deployment guard strict: it still fails when no blog post twin exists.

Validation:

- Reproduced the Astro content sync, build, and Markdown twin generation.
- Confirmed the old check fails and the new check passes on generated output.
- Confirmed the new check fails for an empty blog tree.
- Passed git diff --check, workflow YAML parsing, and the repository pre-commit hook.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -417,7 +417,9 @@ jobs:
           # The agent index lives at the site root (the wave-1/2 subtrees carry
           # their own Markdown twins along with the cp -R above).
           cp next/dist/llms.txt website/build/llms.txt
-          test -f website/build/blog/index.md
+          # Section landing pages have no Markdown source; require a real post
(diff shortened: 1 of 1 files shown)


moonming:
Independent audit completed:

- HIGH: none.
- MEDIUM: the initial find command could match non-regular files and could mask a find error inside test command substitution.
- Resolution: commit fa75e87ae adds -type f and assigns the find result before asserting it, so find errors propagate under set -e and an empty result still fails.
- Verification: positive nested file, empty blog tree, and find-error paths all behave as intended; the repository pre-commit hook also passes.

juzhiyuan (approved):

LiteSun (approved):

PR #2106, opened 2026-08-20
docs: release APISIX v3.18.0

shreemaan-abhishek:
## Summary

- Bump APISIX version from 3.17.0 to 3.18.0 across website config files
- Update release date to 2026-08-20
- Add version 3.18 to the versions list
- Refresh the `mcp-bridge` canonical fixture in `deploy.yml`, which 3.18 invalidates

https://github.com/apache/apisix/blob/release/3.18/CHANGELOG.md#3180

The release date matches the publication time of the source artifacts on the ASF mirror (`https://downloads.apache.org/apisix/3.18.0/`, 2026-08-20), the same convention 3.17.0 followed, and `apache-apisix-3.18.0-src.tgz` is already published there so the download link resolves.

### Why this touches `deploy.yml`

The "Assert canonical contract" step pins a fixture on `mcp-bridge` being present on `next` but absent from the latest release, so the `next` page has to self-canonicalize:

```
# `mcp-bridge` exists on `next`, but not in the latest release. It
# must remain self-canonical instead of pointing at a missing local
# version-less page.
```

`docs/en/latest/plugins/mcp-bridge.md` landed in 3.18 (apache/apisix#13640). Once 3.18 becomes the latest version, `LayoutHead` finds the doc in the latest release and points the `next` page at the version-less URL instead, which is the documented behaviour:

```js
const latestRelease = activePlugin?.versions.find(({ name }) => name !== 'current');
const latestDoc = latestRelease?.docs.find(({ id }) => id === activeDoc?.id);
```

So the fixture's expectation is now the stale half, and the assertion is updated to the version-less canonical. Note that the self-canonical branch is left without a fixture: no doc currently exists on `next` that is missing from 3.18. The comment says so, so it can be restored when one appears.


kayx23 (commented):
The 3.18.0 version bump looks correct (changelog, GitHub release, and ASF source tarball all match `3.18.0` / `2026-08-20`). Lint and the broken-link check are already green.

The failing **Test and Deploy Website** job is caused by this PR, not by unrelated deploy flake. **Build** succeeds; **Assert canonical contract** fails.

That assertion still assumes `mcp-bridge` exists only on `/docs/apisix/next/` and is missing from the latest release. That was true for 3.17. It is no longer true for 3.18 (`docs/en/latest/plugins/mcp-bridge.md` is on `release/3.18`). Once 3.18 is the latest version, the `next` page should canonical to the version-less latest URL, so the current self-canonical grep is expected to fail.

Please update `.github/workflows/deploy.yml` in this PR (or a required follow-up on the same branch) so the `mcp-bridge` fixture matches 3.18 being latest. Skipping the deploy check would leave that contract stale.

kayx23 (approved):

nic-6443 (approved):

